PR #2, opened 2020-02-25
Code refactoring + bug fixes

ydkhatri:
Changes
* Code refactored to allow each artifact parser to have its own file under artifacts folder
* Use `os.path.join()` everywhere to handle windows/linux/mac any kind of path
* Some bug fixes in usagestats
* Some bug fixes in parse3 module
* Changed artifact regex terms to make more accurate (this was missing files/folders before)


aleapp.py
```diff
@@ -12,134 +13,122 @@
...
-			'usagestats':'*/usagestats/*',
-			'recentactivity':'*/system_ce/*'}
-'''
-tosearch = {'redditusers':'*Data/Application/*/Documents/*/accounts/*',
-			'redditchats':'*Data/Application/*/Documents/*/accountData/*/chat/*/chat.sqlite'}
-'''
-	
 os.makedirs(reportfolderbase)
-os.makedirs(reportfolderbase+'Script Logs')
+os.makedirs(os.path.join(reportfolderbase, 'Script Logs'))
```
agamdua: Awesome, this is the right way to handle file paths, I'll create an issue in iLEAPP as well to do the same there.

scripts/artifacts/recentactivity.py
```diff
@@ -0,0 +1,216 @@
...
+                    #else:
+                    #    recimg = 'NO IMAGE'
+                    #insert all items in database
+                    cursor = db.cursor()
+                    datainsert = (task_id, effective_uid, affinity, real_activity, first_active_time, last_active_time, last_time_moved, calling_package, user_id, action, component, snap, recimg, fullat1, fullat2,)
+                    cursor.execute('INSERT INTO recent (task_id, effective_uid, affinity, real_activity, first_active_time, last_active_time, last_time_moved, calling_package, user_id, action, component, snap, recimg, fullat1, fullat2)  VALUES(?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)', datainsert)
+                    db.commit()
+                    
+    #Create html file for report
+    f1= open(os.path.join(report_folder, 'Recent Activity.html'), 'w+')
```
agamdua: I personally prefer using a `with` context manager.

This is ok because the process will eventually exit, so one can trust the OS to clean up the file handles in most cases but using a context manager is the easiest way to not have to deal with that in case execution gets interrupted before you hit `f1.close()`.

scripts/parse3.py
```diff
@@ -451,7 +451,7 @@ def ParseData(data, start, end, messages, depth = 0):
             #start = start + 9
             start = start + 8
             try:
-                floatNum = struct.unpack('d',struct.pack('q',hex(num),16))
```
agamdua: What is the reasoning behind this? I wonder if we need to port the changes to ILEAPP as well since the same file is being used there afaik. Is this not a dependency we can `pip install` so both projects can benefit from bug fixes? I haven't heard of this package before so no idea.

scripts/artifacts/recentactivity.py
abrignoni: Yeah that is my code. This is what I get for trying to put put together old code with new. You are absolutely correct.

agamdua: Cool cool, we'll get around to cleaning it all up eventually.

scripts/parse3.py
ydkhatri: You can't use hex(num) as that returns a string. This is a bug in the library.

ydkhatri: That's a bug in the library. hex(num) returns a string, but it needs a number there.

ydkhatri:
Another thing I did here, is change all tabs to spaces. It's always easier to work with spaces. You might want to change the settings of your editor to insert spaces whenever you hit tab.

scripts/parse3.py
agamdua: yep, `hex()` does return a `str`. makes sense 👍

ydkhatri:
There is more refactoring and cleaning that can be done. Also need to add more error handling at various places.

ydkhatri:
> Hey @ydkhatri! Cool to meet you. @abrignoni may have mentioned I am spending some time on ILEAPP and asked if I could take a quick look here.
> 
> This is a cool PR! Added an issue in ILEAPP ([abrignoni/iLEAPP#10](https://github.com/abrignoni/iLEAPP/issues/10)) as well thanks to this pull request.
> 
> The more contributors on these projects the better it is, well met! :)

Welcome @agamdua ! Good to meet you virtually.

agamdua:
> Another thing I did here, is change all tabs to spaces. It's always easier to work with spaces

100%

Regarding formatting in general, check this out, after you PR is merged I (or you or anyone) can add the same thing to ALEAPP:
https://github.com/abrignoni/iLEAPP/pull/7

PR #3, opened 2020-02-28
bug fixes + improvement

ydkhatri:
Zip works again now!
Changed regex for wellbeing
Now handles activities and usage stats for more than one user (if present)


abrignoni:
I owe you dinner. This is awesome.

Issue #6, opened 2020-03-02
Artifact_Report.py can't html escape int and NULL values


abrignoni:
Got this error:

<img width="1433" alt="Screen Shot 2020-03-02 at 12 42 55 AM" src="https://user-images.githubusercontent.com/28718987/75648669-c8735f80-5c1e-11ea-91cc-39ce12452b5a.png">

Was able to run the report function by using a CASE statement in the SQL query to clean up the null values and str(row[int]) at the data_list.append line. 

Any thoughts on how to normalize the output so no matter the field value fed to artifact_report.py (line 68) escapes them as strings? @ydkhatri

abrignoni:
The other error was:
> AttributeError: 'int' object has no attribute 'replace'
NULL and int. Hope I made sense.

ydkhatri:
This should be an easy fix. In the report function, we can just convert every input sent to a string using `str(..)` I will make the change.

abrignoni:
Worked like a charm!

PR #7, opened 2020-03-02
+ string conversion to write_artifact_data_table

ydkhatri:
Now parsers don't necessarily need to worry about the input they pass to write_artifact_data_table.


ydkhatri:
Fixes issue #6

PR #27, opened 2020-03-22
Adding google quick search and now parser

ydkhatri:
I'm including some protobuf parsing code from [blackboxprotobuf](https://github.com/nccgroup/blackboxprotobuf). This is a better (more complete) implementation for raw protobuf parsing and has more features than the one currently used in some other artifacts.

Also reduced the gui size for aleapp, as it needs to run on smaller screens. A lot of people run their tools in a VM with smaller screens.

### DIFF
--- a/aleappGUI.py
+++ b/aleappGUI.py
@@ -11,17 +11,18 @@ from time import process_time, gmtime, strftime
 sg.theme('LightGreen5')   # Add a touch of color
 # All the stuff inside your window.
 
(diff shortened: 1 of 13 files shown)


abrignoni:
Tested on Android 9 and 10. This artifact is SIIIIIIICK in the best way possible. Love it!
Hope a blog post on it is on the way.

ydkhatri:
> Hope a blog post on it is on the way.

When I get some time..

PR #28, opened 2020-03-22
Google music

abrignoni:
Parses Google Music history. SQLite DB has a protobuf field. Could be parsed more efficiently with the newer protobuf implementation. Currently writing blobs out from DB to disk to then parse. Makes me sad.

### DIFF
--- a/scripts/artifacts/googleMusic.py
+++ b/scripts/artifacts/googleMusic.py
@@ -1,0 +1,73 @@
+import sqlite3
+import textwrap
+import json
+from scripts.parse3 import ParseProto
+
+from scripts.artifact_report import ArtifactHtmlReport
+from scripts.ilapfuncs import logfunc, is_platform_windows
+
+def get_googleMusic(files_found, report_folder, seeker):
+    file_found = str(files_found[0])
+    db = sqlite3.connect(file_found)
+    cursor = db.cursor()
+    cursor.execute('''
+    Select
+    CASE
+        timestamp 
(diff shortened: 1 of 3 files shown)


abrignoni:
Sample test data:
[GoogleMusicTestData.zip](https://github.com/abrignoni/ALEAPP/files/4364790/GoogleMusicTestData.zip)

ydkhatri:
ok, will take a look shortly.

abrignoni:
Going to change the search expression and the names of the artifact. The database does not have the .sqlite extension in the original extraction. This is what happens when I work with provided test data. LOL. Just fyi. Not touching anything else.

ydkhatri:
Got it. Also what version of android or google music is this? I am not seeing this in any of the images I have.

abrignoni:
Changed the name. Is called Playing Now. Got it from Kevin Pagano. Let me
ask him about versions and whatnot.

abrignoni:
Android 10, app is Now Playing, built into the Pixel series starting at v2.
It's part of the Ambient Services found here,
https://play.google.com/store/apps/details?id=com.google.intelligence.sense

ydkhatri:
OK, so its not Google Music at all, but another service that's showing you _any music playing nearby_, according to their description.

abrignoni:
Yes. That is why I changed the name from Music to Playing Now as stated previously. 

Interesting the fact about keeping track of any music playing nearby. Did not know about that.

PR #30, opened 2020-03-24
Changes to now playing

ydkhatri:
- Changed protobuf parser
- Added short description in output file (we should add to other modules too)
- Grouped similar items (reduces output rows by a third)
- Removed protobuf from output as nothing seemingly important there.
- Added album and year


abrignoni:
Having all the timestamps in one field was a great idea!!!!
Also the album year and name.

PR #47, opened 2020-05-03
GUI magic - Single input path + Progress bar

ydkhatri:
There are now 2 buttons to browse folder or file. Selecting the input type dynamically enables/disables them. 

While at it, I also added a progress bar. Enjoy!

image


abrignoni:
This is siiiiick. Love the progress bar and how much more clean selection of data to process is.

PR #55, opened 2020-06-05
Zynga Chess Chat 

mastenp:
I am new to this, so if this is wrong I apologize.  Also want to make sure that Jessica Hyde (@B1N2H3X) gets appropriate credit for the SQL Queries that were taken from her Artifact Exchange .xml.

### DIFF
--- a/scripts/artifacts/zyngachessChats.py
+++ b/scripts/artifacts/zyngachessChats.py
@@ -1,0 +1,50 @@
+import sqlite3
+import textwrap
+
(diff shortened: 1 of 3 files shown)


abrignoni:
Added an icon to the chats category that the artifact will be under.
Please run again @mastenp and make sure the category icon is correct. 
If you could share a screenshot of how the report looks that would be awesome.

mastenp:
ChatIconScreenshot
Looks great.

PR #60, opened 2020-06-16
Add Parsing

mastenp:
Add Parsing for the agent_accounts.db obtained with Magnet ACQUIRE.
Add Parsing for the acquiring_contacts2.db obtained with Magnet ACQUIRE.

### DIFF
--- a/scripts/report.py
+++ b/scripts/report.py
@@ -24,6 +24,7 @@ def get_icon_name(category, artifact):
         else:                           icon = 'user'
     elif category == 'CALL LOGS':       icon = 'phone'
     elif category == 'USER DICTIONARY': icon = 'book'
+    elif category == 'CALENDARS':       icon = 'calendar'
     elif category == 'CHROME':          
         if artifact.find('SEARCH TERMS') >= 0:      icon = 'search'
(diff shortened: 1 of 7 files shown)


scripts/report.py
```diff
@@ -132,7 +134,8 @@ def generate_report(reportfolderbase, time_in_secs, time_HMS, extraction_type, i
     
     elements_folder = os.path.join(reportfolderbase, '_elements')
     os.mkdir(elements_folder)
-    __location__ = os.path.dirname(os.path.abspath(__file__))
+    __location__ = os.path.realpath(os.path.join(os.getcwd(), os.path.dirname(__file__)))
```
ydkhatri, line 137: Any reason for this change? Why is os.getcwd() included? Is there a special case where
`__location__ = os.path.dirname(os.path.abspath(__file__))` does not work?

mastenp: This pull can be closed.  I haven't had time to add WAL File inclusion that Alexis asked for, and I believe there have been many updates since this Pull Request.  Also not sure why that line was changed as it doesn't fall into the small area of code for adding a new artifact.  Thanks for all you and Alexis do though.  I'll resubmit it when I have time for the WAL.

PR #85, opened 2020-12-22
Offline Google docs extraction + minor cleanup

ydkhatri:
Added missing requirement (simplekml) too.


ydkhatri:
Icons for file/folder added too. And if the file is available offline , a link to the extracted file (in green).

image

PR #98, opened 2021-01-11
Add Cast parser

deagler4n6:
Add parsing of cast.db


abrignoni:
This is pretty cool. Thank you so much!!!
If you want to be part of the contributor's list send me a PR of the version_info.py script with your chosen data added to it.

PR #101, opened 2021-01-15
Add 'Files By Google' parsers

stark4n6:
Pulls search history and file master table information


abrignoni:
I dont have test data for these artifacts. Question, are you pulling any possible wal or journal files that might be available with the database?

If so on your regex make sure to pull the wal and journal files. Then use the following pattern (modify as necessary) to open the database. 

Screen Shot 2021-01-15 at 11 44 43 AM

Python will take into account the journaling files if they are present.
Let me know and I'll merge immediately.

Good stuff. So useful to have multiple artifacts showing file presence / metadata.

stark4n6:
From what I saw, there were no WAL or journal files created oddly enough, attached are sample files if you want to test.
[FilesByGoogle.zip](https://github.com/abrignoni/ALEAPP/files/5822870/FilesByGoogle.zip)

abrignoni:
If it is a sqlite database it might have some journaling depending on when
the extraction was made. I am making the habit of taking them into account
on my sqlite artifacts in case the extraction has them.

Plan to update my old sqlite artifacts too.

stark4n6:
My test data didn't have any and neither did Josh's Android 11 image so I'm guessing they don't get created

ydkhatri:
> My test data didn't have any and neither did Josh's Android 11 image so I'm guessing they don't get created

It's always a good idea to account for them, you never know when they might appear there.

stark4n6:
Will close and add the new changes

PR #106, opened 2021-01-25
Runtime permissions per app

abrignoni:
Compatible with Android 10 & 11.
List of app permissions as accepted per the user.


abrignoni:
Artifact pulls app permissions per user.

Screen Shot 2021-01-25 at 11 06 15 AM

PR #115, opened 2021-02-20
Handle ab files

edward-greybeard:
Added ability to handle android backups.
Note; the tar will be placed within the temp directory of output, and will be processed from there as if it was a normal tar.

### DIFF
--- a/aleapp.py
+++ b/aleapp.py
@@ -3,6 +3,7 @@ import os
 import scripts.report as report
 import shutil
 import traceback
+import io
 
 from scripts.search_files import *
 from scripts.ilapfuncs import *
--- a/aleapp.py
+++ b/aleapp.py
@@ -12,7 +12,7 @@ from time import process_time, gmtime, strftime
 
 def main():
     parser = argparse.ArgumentParser(description='ALEAPP: Android Logs, Events, and Protobuf Parser.')
-    parser.add_argument('-t', choices=['fs','tar','zip', 'gz'], required=False, type=str.lower, action="store", help="Input type (fs = extracted to file system folder)")
+    parser.add_argument('-t', choices=['fs','tar','zip', 'gz', 'ab'], required=False, type=str.lower, action="store", help="Input type (fs = extracted to file system folder)")
     parser.add_argument('-o', '--output_path', required=False, action="store", help='Output folder path')
     parser.add_argument('-i', '--input_path', required=False, action="store", help='Path to input file/folder')
     parser.add_argument('-p', '--artifact_paths', required=False, action="store_true", help='Text file list of artifact paths')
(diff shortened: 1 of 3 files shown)


ydkhatri:
Looks good, haven't run it yet. Can you operate on the files within, without writing out the entire tar? 

I've been looking into the backup tar format since 2019 and there are quite a few things that can be parsed. Some databases are same as on disk, others are very different. I have some code [here](https://github.com/ydkhatri/Android/tree/master/adb_backup_parser) that reads data from key value pair backups which quite a few tools missed (at least when I wrote it). I should integrate that code here. There is more on my local pc, not in github.

abrignoni:
Tested with a sample ab file. Creates a tar file I can't open with any tool.

edward-greybeard:
> Looks good, haven't run it yet. Can you operate on the files within, without writing out the entire tar?

Rather than write a new FileSeeker, I thought it would be better to just change it into a tar and then run the FileSeekerTar.

> Tested with a sample ab file. Creates a tar file I can't open with any tool.

While it doesn't open the tar file, does ALEAPP run? Some tools (like 7zip it seems) refuse to open it whereas others don't complain (like Python tarfile). I've checked it against an AB file I've just generated from a phone running Android 11 and if worked fine.

Try running a tar tzvf backup.tar.gz and seeing if it works on your test data.

upintheairsheep:
http://fileformats.archiveteam.org/wiki/Android_ADB_Backup

abrignoni:
Thanks for this, and sorry it sat open this long.

I went through it properly. Two things stop it going in.

First, the tar. The payload of an .ab is a zlib stream, not gzip. Prepending a gzip
header gets you a file whose deflate data happens to start in the right place, and
whose trailer is a 4 byte adler32 where gzip wants 8 bytes of CRC32 and length.
Python's tarfile reads it because it stops at the end of archive marker before it
reaches the trailer, which is why it worked for you. gzip -t and bsdtar both reject
it, which is what I hit when I tested it in 2021. The code also never reads line 3 of
the header, so an uncompressed backup gets a gzip header stapled onto a plain tar, and
the is_tarfile check assigns its result without testing it, so it logs "Output
verified OK" either way.

Second, and this is the bigger one. A backup does not use device paths. It stores
/data/data/com.foo/databases/x as apps/com.foo/db/x, and shared_prefs as sp. Our
artifact paths are written against the device layout. I took a real full filesystem
extraction, relabelled its files the way a backup names them, and ran every declared
pattern against both: 530 artifacts fire on the extraction, about 100 on the backup.
So even with a valid tar the run looks fine and reports almost nothing. It needs a
name mapping in the seeker, and with one it goes back to about 440.

The branch is also more than 3,000 commits behind and all three files conflict. The
GUI was rewritten to tkinter since you wrote this.

Closing it, not because the idea is bad. If an .ab turns up in a case I will build the
converter and the mapping properly. In the meantime android-backup-extractor converts
these files, though the path layout point above still applies to the tar it hands you.

Issue #178, opened 2021-12-16
Missing requirements


kibaffo33:
Hey Briggs. Big fan of the XLEAPP tools, thanks for all your efforts and sharing.

Requirements need updating following SGallery Calculator Vaults addition (https://github.com/abrignoni/ALEAPP/commit/4e44c18ec1fdad0ce0265444dc2b097ba1f86ea5). Standby for PR...

kibaffo33:
PR https://github.com/abrignoni/ALEAPP/pull/179 relates

abrignoni:
What platform are you seeing the error in? Is mimetypes-magic==0.4.30 necessary? I don't remember having to pip install it nor would I know where would it be used. Current code identifies the mimetypes using the magic or filetype libraries. Maybe adding mimetypes-magic==0.4.30 causes some sort of conflict? Not sure what might be going on. Thoughts?

kibaffo33:
Ah, I probably misunderstood which `magic` library you're using. Is it `python-magic-bin`? 

I was on a MacOS 12 (Intel) earlier, but also had the same issue with Windows 10. 
Now I'm on a Mac with M1, which `python-magic-bin` doesn't support.

May have to come back to this on Monday...

kibaffo33:
You were right, `mimetypes-magic` was unnecessary!

I've removed it from https://github.com/abrignoni/ALEAPP/pull/179 and this works now...

kibaffo33:
Thanks for merging!
https://github.com/abrignoni/ALEAPP/commit/d7d3bfbe9f06ae52702e1561b99de570edecf7f4

PR #179, opened 2021-12-16
Adds requirements for SGallery

kibaffo33:
Relates to #178 

Hi Briggs, these requirements appear to be missing following the SGallery addition.

Still appears problematic though... 
- When I install the original `requirements.txt`, followed by the three additions, ALEAPP runs fine.
- When the three additions are inside `requirements.txt` and installed, ALEAPP still missing libraries...

Now I'm confused!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,6 @@ wheel
 simplekml
 xmltodict
 bcrypt==3.2.0
+filetype==1.0.8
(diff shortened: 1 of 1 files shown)


kibaffo33:
Updated as discussed here: https://github.com/abrignoni/ALEAPP/issues/178

PR #202, opened 2022-01-13
Firefox fix on Report

stark4n6:


stark4n6:
Cancelling as I found more issues

PR #232, opened 2022-03-11
Update smsmms.py

Krypterry:
Provide additional friendly names for SMS/MMS message types as per stock Android (Telephony.java).
Determine if the SMS database contains messages from user explicitly blocked numbers on a Samsung device, and include them in the generated report. These messages may have not been read by the recipient, although they may provide significant forensic value in a harassment investigation.


abrignoni:
Very nice. Thanks!

Krypterry:
You're welcome, glad to contribute!

PR #238, opened 2022-04-27
Fix: `.gitignore` for `.DS_Store` file

digitalisx:
Hello, everyone in the community 🙂

In the process of analyzing the source code,
I confirmed that the `.DS_Store` file was included in our Repository.
In the process of removing it,
I also confirmed that `.gitignore` is not clearly defined.


abrignoni:
Thank you. Hadn't noticed.

PR #255, opened 2022-06-17
SnapChat My Eyes Only PIN Bruteforce

InvertReality:
Script to brute-force SnapChat My Eyes Only PIN

Memories.db stores the PIN hash in the 'memories_meo_confidential' table - 'hashed_passcode' column as a bcrypt hash. 

Memories.db will only be available/contain a hashed passcode if the user has entered the MEO PIN code while connected to the internet. During my testing, when logging out of SnapChat, the memories.db database was deleted/no longer accessible.
Memories.db was re-created upon login, but the hashed_passcode column was empty until I entered the MEO PIN code
while connected to the internet.

It seems that once the PIN code is verified online initially, and a bcrypt hash is created and stored offline within Memories.db,
the user may then enter the PIN to access MEO folder offline (the hash will be cached for an undetermined period of time - this may be the same as the 60 day timeout reported on the Forensics Discord server for media).

memories.db/hash is persistent through power off/on states. Logout and x days is the only thing that appears to remove the hash. However, may still be handy if device is seized and examined early.

Example bcrypt hash [PIN - 1234]:  $2a$06$hJJsu8Yb5tlJtsQzMX3ZJu9u6JIiZaoq60U2AbBpx57CImBDrx15q

Iterations = 6
Salt = $2a$06$hJJsu8Yb5tlJtsQzMX3ZJu
Password hash = 9u6JIiZaoq60U2AbBpx57CImBDrx15q

I have some test memories.db so that you can test the script (one db which contains the hash, and one without for error checking via script).

My Discord is AeroDai#1361 and I'm on the Forensic Discord server.

### DIFF
--- a/scripts/artifacts/snapchatMEO.py
+++ b/scripts/artifacts/snapchatMEO.py
@@ -1,0 +1,69 @@
+# snapchatMEO: Brute-force SnapChat My Eyes Only PIN Code
+# Author: David Mann {Twitter: @Sector7Reactor}
+# Date: 17/06/2022
+# Artifact version: 1.0.0
+# Android version tested: 12
+# Requirements: None
+
+import bcrypt
+import sqlite3
+import itertools
+import string
+
+from scripts.artifact_report import ArtifactHtmlReport
+from scripts.ilapfuncs import logfunc, tsv, timeline, is_platform_windows, media_to_html, open_sqlite_db_readonly
+
+
+def bruteforceMEO(encrypted_password):
+
+    for numbers in itertools.product(string.digits, repeat=4):
+
+        pin = ''.join(numbers)
+        check_hash_match = bcrypt.checkpw(pin.encode('utf-8'), encrypted_password.encode('utf-8'))
+
+        if check_hash_match == True:
+            logfunc(f'PIN code match found: {pin}')
+
+            return encrypted_password, pin
+    else:
+        return None
+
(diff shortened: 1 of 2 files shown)


abrignoni:
Thanks for the contribution!

- `snapchatMEO.py`: please include a fixture with this PR.

**Adding a fixture**

Size rules:

abrignoni:
Thanks for this, David, and sorry for the long delay in getting back to you.

Snapchat My Eyes Only recovery is already in ALEAPP. It lives in `snapchat.py` as the "Snapchat - MEO My Eyes Only" artifact (`get_snapchat_meo`) and covers the same ground plus a little more: it reads `user_id`, `hashed_passcode`, `master_key` and `master_key_iv` from `memories_meo_confidential`, recovers the 4-digit passcode with bcrypt, stages the WAL sidecars through the `memories.db*` glob, and ships in the current `__artifacts_v2__` format with sample data across several test images. With that overlap, and since this PR now conflicts with the current tree (`ilap_artifacts.py` no longer exists), I'm closing it as superseded.

The field notes in your write-up are the part worth keeping. That the hash only shows up in `memories.db` after the PIN is entered while online, that the database is removed on logout and returns empty until the PIN is re-entered, and that the hash then survives reboots for a bounded window are useful for examiners and all point to imaging early. I'll look at adding a short version of that to the artifact's notes.

Thanks again for documenting it and offering test data. Keep the contributions coming.

PR #286, opened 2022-08-17
Improved plugins

cclgroupltd:
Allows plugins to mark themselves as required or long-running (the latter currently has no impact on how they are used).
Default plugins are always selected - this has allowed us to purge the hard-coded usagestatsversion mentions as the plugin just marks itself as required.

Additional fix - users are now alerted to missing plugins in profiles


jijames:
This is a much more elegant solution. Thanks!

PR #289, opened 2022-09-06
[protonVPN.py] - Included VPN Connection History

nxb1t:
VPN Server IP is logged in : /data/data/ch.protonvpn.android/log/Data.log


abrignoni:
Do you close the file that is open at line 69 after you are done with it?

PR #297, opened 2022-11-03
Update FacebookMessenger.py

theAtropos4n6:
I noticed that from at least version 379.1.0.23.114 onwards the app changed the name of the db from threads_db2 to ssus.USER-ID.threads_db2. Hence the update. The parser seems to still work just fine. However further research may be needed though to verify if anything else changed.


stark4n6:
Thanks for pointing this out, I'll have to see if I can get my test phone back up and running and check out any further changes.

theAtropos4n6:
That would be awesome. Thanks Kevin!

PR #339, opened 2023-02-28
Garmin Dailies API

fabian-nunes:
Decided to add another complementary module for the Garmin API (display the general daily data recovered from the API, the script for that is already updated in our other repo). This view is the same as the daily summaries retrieved from the private repository.


abrignoni:
Been trying to make a binary using pyinstaller using a spec file (aleappGUI.spec) and keep getting this error.
Any thoughts on what it might be? I have added the new files your artifacts require to the spec file. Not sure why I get these error when I run the generated binary.

Capture
Capture2

abrignoni:
I think i was editing the wrong spec file. Never mind

PR #355, opened 2023-03-27
Fixed Bug nike artifact

fabian-nunes:
Fixed bug where macos would confuse database file with journal file, did the same fix for other modules in case of a possible bug.

### DIFF
--- a/scripts/artifacts/AdidasUser.py
+++ b/scripts/artifacts/AdidasUser.py
@@ -95,6 +95,6 @@ def get_adidas_user(files_found, report_folder, seeker, wrap_text):
 __artifacts__ = {
     "AdidasUser": (
         "Adidas-Running",
-        ('*com.runtastic.android/databases/user.db*'),
+        ('*com.runtastic.android/databases/user.db'),
         get_adidas_user)
 }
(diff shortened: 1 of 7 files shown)


abrignoni:
I recommend a different fix. By not pulling in the wal or journal files at grep, processing will show incomplete data. Go to the beginning of your artifact function and change the initial loop as so:

`for file_found in files_found:
        file_found = str(file_found)
        if file_found.endswith('name_of_the_db' ):
               #Do the things `

They way described above will take into consideration the journal & wal file content when the database is connected to.
The same code can be expressed in the negative.

`for file_found in files_found:
        file_found = str(file_found)
        if not file_found.endswith('name_of_the_db'):
            continue # Skip all other files`
  
I saw on your code kinda the same idea by trying to limit files_found to the file you need. In some artifacts files_found will have the database files but also other files that you might need. The suggested code above should address it. I never alter files_found. I pull from it as needed.

fabian-nunes:
Alright thanks for the information, i am still solving another bug I found in the artifacts, so I will probably close this request and then make another one, the problem is that the files don't follow a normal convention (eg db, db.wal, db.smh) instead they are called db, db-journal and it caused a conflict, the problem only happened in my mac not my windows, so I will try and test your method.

PR #377, opened 2023-05-01
Create PythonHistory

upintheairsheep2:
This is actually an alternate account I created after being mistakenly flagged as a bot by GitHub. I apologize for my previous behavior of requesting you to parse every application ever created without providing any sample data or code to assist you.

### DIFF
--- a/scripts/artifacts/PythonHistory
+++ b/scripts/artifacts/PythonHistory
@@ -1,0 +1,33 @@
+import codecs
+import csv
+from scripts.artifact_report import ArtifactHtmlReport
+from scripts.ilapfuncs import logfunc, tsv, is_platform_windows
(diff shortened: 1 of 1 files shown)


abrignoni:
Thank you for the idea here. Interactive interpreter history is genuinely probative when it exists, so the target is a fair one. Two things stop this from landing: the file was committed without a .py extension so the plugin loader never picks it up, and none of our test images carry .python_history, shell history or Termux at all, so there is nothing to validate a parser against. I checked all twelve Android corpora, including the tar image, and found no shell history of any kind. I am recording Termux and shell-history coverage on our test-data wish list, and if an image carrying it turns up this becomes straightforward to build properly against real data. Closing for now, with appreciation.

PR #379, opened 2023-05-03
Add badoo modules

fabian-nunes:
Create two modules for the dating app Badoo:

The **chat module** will extract all the conversations between users: (text, video, audio, images etc). The module also extracts connection information, such as name, gender, work, and all his photos.

<a href="https://ibb.co/Kh5Myq1">Captura-de-ecr-2023-05-03-s-20-53-17</a>

To turn this into the most realistic possible, we added a new feature to ALEAPP. We present the chat window:

Captura-de-ecr-2023-05-03-s-20-38-58

By clicking on the button Open chat in the datable related to the connection, the report will open the conversation that happened in a chat window format for better understanding.

This feature is purely CSS and JS and can be adapted to other applications that store conversations.

Another module created is related to **notifications of the connections**, such as:
Persons that viewed your profile, messages sent without connection, etc.
This module gives the message, the user's name, and the profile picture.

There is room for improvement. Since I used fake accounts created with temp mail it was quite challenging to generate data since Badoo is one of the few apps that requires photo recognition to detect false accounts, I managed to bypass this, but I only had minutes for it.

### DIFF
--- a/scripts/artifact_report.py
+++ b/scripts/artifact_report.py
@@ -339,3 +339,23 @@ class ArtifactHtmlReport:
             }});
             </script>
             """
+
+    # Function to add a chat window to the artifact
+    def add_chat(self):
+        self.report_file.write('<div class="container py-5">')
+        self.report_file.write('<div class="row d-flex justify-content-center">')
+        self.report_file.write('<div class="col-md-8 col-lg-6 col-xl-4">')
+        self.report_file.write('<div class="card" id="chat">')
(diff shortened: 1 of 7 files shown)


fabian-nunes:
@abrignoni is everything in order with the request or do i need to change something?

abrignoni:
All is good. Haven't had a chance to test it out. Will do so shortly. My
bad.

PR #387, opened 2023-05-26
Modification of GPS fitness artifacts

fabian-nunes:
Modification of Fitness modules related to GPS tracking of the apps:

- Adidas
- Garmin
- Nike
- Runkeeper
- Map My Walk
- Strava
- Puma

A new file is generated for each record, an XLSX file containing the coordinates, possibly the timestamp when the user was there, and the respective **road**, **city**, **postcode**, and **country**. This is archived using the Python dependencies **geopy** that returns this information from the coordinates.

To not overload this library and create a cache of information, the data is stored in a local SQLite3 database. In future generated reports, the users ALEAPP will get the data from the database instead of the library.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,6 @@ ENV/
 
 # OSX Cache
 .DS_Store
+
+#GPS database cache
+coordinates.db
--- a/aleappGUI.py
+++ b/aleappGUI.py
@@ -6,7 +6,6 @@ import PySimpleGUI as sg
 import webbrowser
 import plugin_loader
 from scripts.ilapfuncs import *
-import scripts.ilapfuncs
 from scripts.version_info import aleapp_version
 from time import process_time, gmtime, strftime
 from scripts.search_files import *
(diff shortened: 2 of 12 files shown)


fabian-nunes:
Added a new commit. There may be people who do not have access during analysis or do not want to use network functionalities (get data from API and libraries), so I created a simple trigger button to activate or disable these features. By default, it will be disabled, and the modules will work just as before by activating the features of the commit before they are used (generate Excel with data). This can be used for any new module that uses features with network access.

<a href="https://ibb.co/NNmz1Sp">Captura-de-ecr-2023-05-29-s-10-48-25</a>

abrignoni:
Instead of creating a button to disable some artifacts consider adding those artifacts to the unchecked by default list in aleappGUI.py as seen in the attached image. Thanks!

Screenshot 2023-05-29 at 7 29 02 AM

fabian-nunes:
Thanks for the response. The idea of the button is not to disable the modules but to disable the specific network function of the modules (without a network, you still can generate the coordinates maps and kml file, just not the excel). 
I had different ideas, for example, a pop-up when selecting a specific module. However, this form was the simplest for this specific need.

I can just let the modules run as default and see if the user has an internet connection, and if they don't, just not execute the specific part of the module.

abrignoni:
Yes to let them run and if no connection then no execution those parts. 👏

PR #442, opened 2023-12-13
    Fixing the "locate at" field in the html report

segumarc:
Fixing the "locate at" field in the html report for the fitbit and googlemapsgmm artifacts.
The field was being populated with the last file_found read so it was the same everywhere even if the artifacts was from an other.
 
Ex : The Google Maps Label Places report was showing "Google Maps Label Places located at: z:\temp\data\data\com.google.android.apps.maps\databases\gmm_storage.db" and the artifcat came from "z:\temp\data\data\com.google.android.apps.maps\databases\gmm_myplaces.db"


abrignoni:
Great catch! Thanks! 💖

Issue #454, opened 2024-02-01
gmailEmails only process first bigTopDataDB found


patrickdalla:
I processed some evidences with multiple bigTopDataDB sqlite databases. Only the first is processed by current ALeapp plugin.

patrickdalla:
PR #455 would close this

PR #455, opened 2024-02-01
Support multiple Gmail database processing

patrickdalla:
Support multiple Gmail database processing. Closes issue #454


stark4n6:
@patrickdalla I would say more than 2/3 of all the parsers don't support multi-profile at this time. I've been slowly going through and attempting to update them when I have time (and to also support the timezone offsets as well for reports)

PR #487, opened 2024-05-09
Contacts from com.samsung.android.providers

patrickdalla:
In some cases the contacts database is found in the folder pattern */com.samsung.android.providers.contacts/databases


stark4n6:
@patrickdalla are the DB's the same? do they have the same tables and columns? if not the queries may need altered

stark4n6:
@patrickdalla okay will merge

PR #495, opened 2024-05-21
Hashes for artifacts

falcon217836:
Referencing issue  #457, this submission will generate hash values for all artifacts processed.

### DIFF
--- a/scripts/artifacts/hashvals.py
+++ b/scripts/artifacts/hashvals.py
@@ -5,18 +5,17 @@ from time import gmtime
 from time import localtime
 from time import strftime
 from scripts.artifact_report import ArtifactHtmlReport
-from scripts.ilapfuncs import logfunc, tsv, timeline, is_platform_windows, open_sqlite_db_readonly
 from datetime import datetime
 
 
-
 # Write digest results to file
+
 
 def write_hashes(files_found, report_folder, seeker, wrap_text, time_offset):
     sha2pulls = {}
     data_list = []
     for file_found in files_found:
-        artifact_path = file_found.replace(seeker.directory, '')
+        artifact_path = file_found.replace(seeker.directory, "")
         if os.path.isfile(file_found):
             with open(file_found, "rb") as f:
                 fdata = f.read()
@@ -24,23 +23,17 @@ def write_hashes(files_found, report_folder, seeker, wrap_text, time_offset):
 
         else:
             pass
-            
 
         data_list.append((file_found, sha2pulls))
-        
-    report = ArtifactHtmlReport('SHA256 File Hashes')
-    report.start_artifact_report(report_folder, 'Message Digest for Processed Files')
+
+    report = ArtifactHtmlReport("SHA256 File Hashes")
+    report.start_artifact_report(report_folder, "Message Digest for Processed Files")
     report.add_script()
-    data_headers = ['Artifact', 'SHA256']
-    report.write_artifact_data_table(data_headers, data_list, artifact_path, html_escape=False)
+    data_headers = ["Artifact", "SHA256"]
+    report.write_artifact_data_table(
+        data_headers, data_list, artifact_path, html_escape=False
+    )
     report.end_artifact_report()
 
 
-
-__artifacts__ = {
-"File Hashes": (
-    "File Hashes",
-    ('*/*'),
-    write_hashes
-)
-}
+__artifacts__ = {"File Hashes": ("File Hashes", ("*/*"), write_hashes)}
--- a/scripts/artifacts/hashvals.py
+++ b/scripts/artifacts/hashvals.py
@@ -37,3 +37,4 @@ def write_hashes(files_found, report_folder, seeker, wrap_text, time_offset):
 
 
 __artifacts__ = {"File Hashes": ("File Hashes", ("*/*"), write_hashes)}
+
--- a/scripts/artifacts/hashvals.py
+++ b/scripts/artifacts/hashvals.py
@@ -1,0 +1,46 @@
+import os, hashlib, os
+
+from pathlib import Path
+from time import gmtime
+from time import localtime
+from time import strftime
+from scripts.artifact_report import ArtifactHtmlReport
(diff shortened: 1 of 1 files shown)


stark4n6:
@falcon217836 this appears to ingest all files in a path/container and hashes, is that correct? If so how are the speeds? I would imagine that would take a bit of time

falcon217836:
That is correct. And processing was much quicker than I expected. 30K files processed in:

Fs: 5-15 sec
Zip: 10-20 sec
Tar: 10-20 sec
gz: 10-20 sec

There was an additional 15-20 sec required when loading the html report for the artifacts hashed. All in all not too bad for hashing 30K files.

stark4n6:
@falcon217836 I ran it on a test CTF image from MVS2024 and it took 22 minutes just with this parser turned on. I got 69k entries for results with over thousands erroring out 
image

I don't know if this is optimal on a fully utilized FFS extraction. Also the temp folder is almost 22gb

falcon217836:
It wasn't nearly that exhaustive when I ran it in Ubuntu2204, but this brings up something I hadn't considered. Many use ALEAPP on Windows systems so let me make some adjustments, run some regression testing in a Winenv, and I'll update you when I've got a more efficient commit.

JamesHabben:
moving to draft for now. where are we at with this? can the speed be improved to have minimal impact? if not but we still want to merge this in, we probably need to gate it behind a cmd line parameter or GUI check to enable - off by default

abrignoni:
Thanks for the contribution!

- `hashvals.py`: please include a fixture with this PR.

**Adding a fixture**

Size rules:

abrignoni:
Thanks for this, @falcon217836, and apologies it has sat this long.

I ran the PR against current main (`56efd1f6`) to see where it stands. Two things stop it before it hashes anything, and both of them are the framework moving under you since May 2024, not something you got wrong.

**1. The signature.** At your merge base, `aleapp.py` called artifacts with five arguments, `time_offset` included. It passes four now:

```
Reading File Hashes artifact had errors!
Error was write_hashes() missing 1 required positional argument: 'time_offset'
  File "aleapp.py", line 453, in crunch_artifacts
```

**2. `seeker.directory`.** At your merge base every seeker set `self.directory`. Only `FileSeekerDir` does today, so with the signature patched, zip input stops at line 18:

```
Error was 'FileSeekerZip' object has no attribute 'directory'
  File "scripts/artifacts/hashvals.py", line 18, in write_hashes
    artifact_path = file_found.replace(seeker.directory, "")
```

The main loop catches both, so the run still prints "Report generation Completed" and the module is just marked Error.

## What the table looks like once it runs

I patched the signature and ran it on a four file test extraction, folder input. Nine matched paths, nine rows:

```
<report>/data/data                              {}
<report>/data/data/data                         {}
<report>/data/data/data/com.probe.app           {}
<report>/data/.../com.probe.app/databases       {}
<report>/data/.../databases/b.bin               2a747945c3245ef3a6ad49a8166eb281def0af4c38635a1b2ec657d523011556
<report>/data/.../com.probe.app/files           2a747945c3245ef3a6ad49a8166eb281def0af4c38635a1b2ec657d523011556
<report>/data/.../files/c.txt                   480c2336b410f1ad5f8bf1b28944490255804b65350c527787e74ebdd511e3a4
<report>/data/.../files/a.txt                   cbbdbb59c624f12b58fcb63612b5e923b2fed5f42e8827f2d291c760f8fac18f
<report>/data/.../files/link_to_a.txt           cbbdbb59c624f12b58fcb63612b5e923b2fed5f42e8827f2d291c760f8fac18f
```

Three things there matter:

- **A directory is given the previous file's hash.** `sha2pulls` is only assigned inside the `os.path.isfile` branch, so the bare `else: pass` leaves the last value in place. The `files` directory row above carries `b.bin`'s real SHA256, which I confirmed with `shasum -a 256`. A wrong hash on a report row is the thing this artifact cannot ship with, because it reads exactly like a right one.
- **Before any file is read the column prints `{}`**, the dict the variable starts as.
- **The digest is of the staged copy, so it moves with the input type.** The same symlink staged as 12 bytes from the folder input, its target's content, and as a 5 byte file holding the text `a.txt` from a zip. One evidence item, two digests depending on `-t`.

Separately, each row carries the full staged path off my machine instead of the evidence path. The relative path is computed into `artifact_path` and then not used in the row.

## On the speed question

@stark4n6's 22 minutes and 22 GB still holds, and the images have grown since. The `*/*` pattern matches everything, and the seeker stages every match before the module is called, so the copy is paid whether or not the hashing works. Counting members off the central directory of four Android full filesystem extractions here: 44,835, 82,679, 124,949 and 216,064 paths, staging 16 to 32 GiB each.

Two things have changed that also work against the output, @JamesHabben:

- HTML tables over 50,000 rows are now held back from the page, with a notice sending the reader to the LAVA database and the TSV export. This module writes neither, so on a full extraction the hashes would land nowhere at all.
- The artifact block is the v1 `__artifacts__` form. The loader still takes it, but there are none left in the repo, so this would be the only one. Everything else is `__artifacts_v2__` with `@artifact_processor` returning `(headers, data_list, source_path)`.

## Where I think this belongs

The idea in #457 is one I want. I do not think it can live in an artifact module, though. An artifact only sees what the seeker already staged, and making it see everything means copying the whole extraction, which is the cost we keep running into.

The seeker layer is already most of the way there. Every staged file is recorded with its evidence relative path in the LAVA file path list, and the seekers already compare content when they have to: `FileSeekerTar._member_content_key` SHA256s a member's bytes, and the zip seeker uses the recorded CRC-32 and size. Both of them already report type and link target instead of a digest for anything that is not a regular file, which is the distinction that bit the table above. Hashing at staging, for the files the artifacts actually matched, costs very little on top of the copy we are doing anyway and puts the value into LAVA, the TSV and the HTML in one place.

If you want to take that on I will happily point you at the spots. If you would rather leave it here, that is completely fine and thank you for raising it either way. The issue stays open, and this PR is a useful record of what the direct version costs.

One last note if you do pick it up: lint fails on the file now, 7 new warnings, 2 unused arguments and 5 unused imports.

PR #504, opened 2024-07-09
updated crunch artifacts function. It should process tar files faster…

zacharyhays9:
This code will enhance the speed of processing. It will work for tar files but not for .zip and .fs files. I wanted to send this in the prototype version to verify that that this reworking is acceptable. Please let me know what you think! If someone can send me zip and fs archives I will gladly make them work for this code.

### DIFF
--- a/aleapp.py
+++ b/aleapp.py
@@ -9,6 +9,7 @@ import traceback
 import tarfile
 import fnmatch
 import zipfile
+import shutil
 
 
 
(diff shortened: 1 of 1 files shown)


Johann-PLW:
Hi @zacharyhays9,
I have tested your prototype version and at the moment, it does not work as expected.
On my M1 Max Mac Studio, with Josh's Android 12 public image (Android 12 - Data.tar) and all modules selected except walStrings:
- The processing time took:
  - 04:40 in GUI and 03:59 in CLI with the actual crunch_artifacts function.
  - 34:57 in GUI and 01:10 in CLI with your code. Scroll bar does not work anymore in GUI.

In both CLI and GUI, any artifact was decoded (The Processed Files list is empty).

zacharyhays9:
Thank you for your feedback! I will work on the necessary changes.

zacharyhays9:
Wanted to make sure you saw the new commit @Johann-PLW. Hopefully this runs better for you!

Johann-PLW:
Hi @zacharyhays9,
Sorry for the late reply. I saw your new commit but I wasn't able to test the changes as I don't have access to my Mac at the moment but only to my iPAD.
I'll get back to you as soon as I've tested your update.

PR #513, opened 2024-08-14
Add support for reading *sms_backup and *mms_backup files

ydkhatri:
The android backup (backup.ab) now uses this format for storing sms and mms data. 
Files are stored under:
/apps/com.android.providers.telephony/d_f/
as 000000_sms_backup , 000001_sms_backup, ...

Each file stores upto 1000 messages. MMS attachments are not backed up, only text is.

Also fixed a couple of compiler warnings for Python 3.12


abrignoni:
Thank you so much @ydkhatri 
Hope all is going well! 👏🎉👍

Issue #514, opened 2024-08-21
Feature request


PensiveHike:
Viewing reports created by the leapps can be slow for users not connected to the internet as the reports are trying to reach CSS stylesheets stored online. 

Could a feature be added whereby a user can select online/offline reports at report creation?
The offline reports would contain no reference to online resources and so would be quicker to load.

I did start looking into this but didn't want to overstep my place by changing the interface. Plus, I'm a simple scripter and there is definitely someone that could achieve better results than me

PR #531, opened 2024-12-01
Added ability to create offline reports

PensiveHike:
Added section in GUI to allow user to select Online/Offline mode. Also added option within CLI.
Offline version does not contain references to css URLs, which can slow down the loading of html files on standalone workstations.


Johann-PLW:
Thank you @PensiveHike.
I remember that this is something you asked last summer.
I have implemented these changes in iLEAPP based on your issue #514 when we modified report structure but not in ALEAPP yet. 
I think that offline report should not be an option but the default behavior of ALEAPP because the majority of users use it offline.
I can't merge these changes as is because the button is cropped and not accessible in the GUI.
image

Let me know if you can modify your changes to generate offline report by default.
Feel free to have a look to this commit on iLEAPP repo: https://github.com/abrignoni/iLEAPP/commit/8820708366e32c59e89398958ffcf793fa41a823
I remember I had to add offline fonts as well.
Let me know if I forgot something or made something wrong. 

Thanks again for your contribution

PensiveHike:
Thank you for reviewing my code. 
The combobox worked in my version of Windows, but I didn't know it would appear differently on other OS's.
If the end report will be offline by default, then I will remove it. 
I'll have a look at your code to see your formatting and fonts used. I'll aim to get this done in the next few days

Johann-PLW:
No rush,

Thank you very much for taking the time to make these improvements.

PR #532, opened 2024-12-03
Google Keep Notes

charpy4n6:
An update to Google Keep Notes is causing only the title of the Google Keep Notes to show in Cellebrite and Axiom. The current Google Keep Notes in ALEAPP parsed no notes from my test data. Updated script to handle new version of Google Keep Notes.


stark4n6:
There is a Google Keep Notes parser already, wondering if there can be a check in place to merge the two and pivot to a different query depending on a table name or some other unique identifier

PR #533, opened 2024-12-09
Changed Reports to be Offline

PensiveHike:
Removed mention of accessing CSS files from the internet within the html report as this would cause a delay in loading times. 
I originally, created an option for a user to choose the output to be online or offline, but from advice given by @Johann-PLW, I made it offline by default with no option for online version. 
Additionally, I copied code from iLEAPP that places files needed for the final report to be in the _elements folder as this tidies the ALEAPP folder, and now only the elements folder is copied to final report folder as opposed to several individual files.


PensiveHike:
I don't really know how Git works, but after generating the previous pull request, I have added additional features. It would appear these new features have been added to the previous pull request, rather than the need to create a new request. The changes for the new features are as follows:

Identified new way to decode encoded strings found within com.sec.android.app.myfiles OperationHistory.db and com.sec.android.gallery3d local.db. These provide greater insight into user interactions. Have implemented this decoding within script titled gallery3DLog.py and have updated smyfilesOpHistory.
Added another entry to decode dictionary within smyfilesOpHistory.
Added icon_mappings for the MyFiles scripts where there were none.
Updated galleryTrash to use latest version of artefacts introduction. Added this to Gallery3D category, so now shows alongside records for Gallery3D Log results.

Johann-PLW:
Hi @PensiveHike ,
I did not push your PR as there is a bug with the ascending/descending symbol in the column headers.
image

PensiveHike:
As I mentioned previously, I am not familiar with Github. Making changes,and then reversing thesegoing back on it to then do further work made the chain messy. Therefore, I cloned Brig's ALEAPP repository again and made the changes on this fresh copy. I have tried to overwrite my repository with this cleaner version and it appears to have worked, but perhaps you can check this?

I found the filter function wasn't working due to css links, so this has been amended.

PR #551, opened 2025-04-18
OneDrive Metadata Artifact

dabeersboys:
Created a OneDrive Metadata artifact based on Anthony Reince's inital work. 

- Artifact could use some work with opening thumbnails in a new tab, but the ability to right-click>> Open Image in New Tab works just fine.

Updated Version info with Anthony Reince info with his permission.


stark4n6:
@dabeersboys had to do some minor fixes for paths and commas but should work for now. I will continue to work on future updates for it in the next week.

PR #558, opened 2025-05-09
Samsung CMHProvider parser

scrptkd:
Parses data from Samsung's CMHProvider database cmh.db.

### DIFF
--- a/scripts/artifacts/cmh_tags.py
+++ b/scripts/artifacts/cmh_tags.py
@@ -54,6 +54,7 @@ def get_cmh_tags(files_found, report_folder, seeker, wrap_text):
     LEFT JOIN tag_map tm ON f._id = tm.fk_file_id
     LEFT JOIN tags t ON tm.fk_tag_id = t.tags_id
     LEFT JOIN ocr_tag ot ON f._id = ot.fk_file_id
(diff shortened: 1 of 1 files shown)


scrptkd:
@stark4n6 thanks for the review. I was aware of the other cmh.db parser. It is mainly focused on geodata. I am only parsing the tags that mark the media files. I've put it on the same category though. 

- For the -wals I noticed on every test I made that no further data are being saved on wal file. A picture deletion would directly have effect on the main .db file. But I can try parsing it if this is the best practice.
- I will take care of the lower/proper case as soon as possible.
- I wasn't aware of that. I will change the order and place datetime taken first.
- I have no previous experience and haven't researched about LAVA output. I could use some documentation or point me on another sample script.

abrignoni:
Thank you for this one. The CMH tag tables are a genuinely useful source and none of it was covered before. I reworked it and merged it as #1090 with you credited as co-author: split into four focused artifacts (scene tags, image OCR text, QR/barcode content, and user tags) so a file with many tags no longer multiplies into duplicate rows, moved it onto the current artifact API with typed timestamps, and validated it against the six of our Android images that carry cmh.db. It surfaces real content, including image OCR text and, on one image, an Olvid invitation with a contact name pulled from a QR code. Closing this in favor of #1090, with appreciation.

PR #573, opened 2025-07-07
Verification of FFS by paths

abrignoni:
Artifact that checks if a list of expected paths exists in an extraction.


JamesHabben:
this is in conflict now since its old. were you intending to merge this in or just create the module for the examiners you were working with?

abrignoni:
Closing without merging. Not being taken forward.

PR #582, opened 2025-07-27
Added Speedtest artifact extractor

its5Q:
This PR implements a plugin to extract artifacts from the Speedtest app, which contain a lot of useful data such as location data and network names. Inspired by the recently finished Belkasoft CTF :)

Test data from the CTF image: [org.zwanoo.android.speedtest.zip](https://github.com/user-attachments/files/21455127/org.zwanoo.android.speedtest.zip)

Resulting report:
chrome_cVDfOxhYH6
image

### DIFF
--- a/scripts/artifacts/speedtest.py
+++ b/scripts/artifacts/speedtest.py
@@ -9,7 +9,6 @@ __artifacts_v2__ = {
         "category": "Speedtest",
         "notes": "",
         "paths": ('*/org.zwanoo.android.speedtest/databases/AmplifyDatastore.db',),
-        "function": "extract_speedtest_test_results",
         "artifact_icon": "loader"
     },
 
@@ -23,7 +22,6 @@ __artifacts_v2__ = {
         "category": "Speedtest",
         "notes": "",
         "paths": ('*/org.zwanoo.android.speedtest/databases/speedtest',),
-        "function": "extract_speedtest_reports_location",
         "artifact_icon": "map-pin"
     },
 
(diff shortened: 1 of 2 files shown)


its5Q:
Looks good now
EmEditor_y6Z4Bnu24J
EmEditor_GmksgmVLvm

stark4n6:
@its5Q do you want to try doing LAVA compliance reports? You can read about docs here:
[https://github.com/abrignoni/iLEAPP/blob/main/admin/docs/module_updates.md](https://github.com/abrignoni/iLEAPP/blob/main/admin/docs/module_updates.md)
[https://github.com/abrignoni/iLEAPP/blob/main/admin/docs/module_updates_advanced.md](https://github.com/abrignoni/iLEAPP/blob/main/admin/docs/module_updates_advanced.md)

its5Q:
Hm, yeah, sure. I just went with the manual report generation because I wanted to merge two tables into one HTML report, but I guess it's fine to split that up and switch to automatic report generation for this module as well. I'll get on that probably tomorrow.

abrignoni:
Here is a video on how to make LAVA compliant artifacts:
https://www.youtube.com/live/qTgZUh4GPxk?feature=shared

its5Q:
@abrigoni I've switched to artifact_processor and renamed some headers so the KML generation works for location data. Also, I've made a tiny change in the OS version module to strip the newline, before that, a trailing newline would end up in the "Build version" field:
EmEditor_3Asa3tqllk

its5Q:
Yeah, that works. Also made some changes to output all timestamps in UTC.

PR #618, opened 2025-11-22
Fix undefined `time_offset` causing NameError in Turbo_Battery timestamp conversion

hipzi:
Resolves a `NameError` in the Turbo_Battery artifact caused by the use of an undefined variable, `time_offset`, during timestamp conversion. The current code assumes that `time_offset` is always provided by the framework, but this artifact does not receive that parameter, leading to a crash.


JamesHabben:
closing since this has been address in modernizing the module with timezone aware timestamps.

PR #619, opened 2025-11-22
Fix missing `saved_status` column handling in Google Duo Notes artifact

hipzi:
Resolves a crash in the Google Duo Notes artifact caused by the use of a column `saved_status` that does not exist in all versions of the `tachyon.db` database.


abrignoni:
Thank you for this. Google Duo's absent saved_status column is now handled on main generically through null_absent_columns, so the notes artifact reads across schema generations without a per-column probe. Closing as superseded, with appreciation.

PR #622, opened 2025-11-25
Fix Error in the ParseFromString() function. The error message was: Unexpected end-group tag.

liizzasulistio:
Fix error if protobuf not found, since Notification History is introduced in Android 11 (if the Android ver. 10 and below)


abrignoni:
Thank you for this. notificationHistory on main now wraps ParseFromString so a file that is not a valid protobuf is logged and skipped rather than crashing the artifact, which covers the failure this addressed. Closing as superseded, with appreciation.

PR #623, opened 2025-11-26
Document known issues for v3.4.0 Windows executable

### LINKED ISSUE #572: Windows build has errors
I did not test older versions, but current 3.4.0 windows exe crashes due to missing xmltodict. I installed xmltodict on the host-python, but it seems that ALEAPP uses an own virtual environment or similar so this does not solve the issue. 
donwloading the zip and running via host-python (after installing requirements) seems to work fine.

Thanks

Gnanasaikiran:
Added known issues section for v3.4.0 Windows executable regarding missing dependency.

Fixes #572

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,17 @@ def get_cool_data1(files_found, report_folder, seeker, wrap_text):
 
 ```
 
+## Known Issues
+
+### v3.4.0 Windows Executable
+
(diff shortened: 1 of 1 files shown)


abrignoni:
Thank you for documenting this. It describes known issues in the v3.4.0 Windows executable, which is several releases behind, so the content no longer matches what ships. Closing as stale, with appreciation.

PR #624, opened 2025-11-26
Improve media file lookup speed (Fixes #591)

### LINKED ISSUE #591: Media Context Improvement
See PR from iLEAPP

https://github.com/abrignoni/iLEAPP/commit/047ac2a90c5de49b4c26cbd90a29fe186f34d299

Gnanasaikiran:
Replaces the slow, linear search in `check_in_media` with a high-performance, two-stage lookup using a pre-computed dictionary.

The previous O(N) search caused severe slowdowns on artifacts with many files. This is now an O(k) operation using a lazily-built lookup map that caches filename-to-path mappings.

Changes:
- Added module-level filename lookup map cache
- Added `_build_filename_lookup_map()` helper function
- Added `_get_filename_lookup_map()` for lazy cache initialization
- Added `get_source_file_path()` API for high-speed file path lookup
- Modified `check_in_media()` to use the new optimized lookup

Ported from iLEAPP commit 047ac2a.

### DIFF
--- a/scripts/ilapfuncs.py
+++ b/scripts/ilapfuncs.py
@@ -152,9 +152,65 @@ def set_media_references(media_ref_id, media_id, artifact_info, name, media_path
     ))
     lava_insert_sqlite_media_references(media_references)
 
+    # Module-level cache for filename lookup map (improves check_in_media performance)
+_filename_lookup_map = None
+_cached_files_found_id = None
+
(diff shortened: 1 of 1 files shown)


JamesHabben:
this PR got a little stale and the framework has been refactored a couple times. this is no longer the structure since this part of the code has now moved to the context class. closing the pr.

PR #626, opened 2025-11-28
Fix Resolve crashes in Turbo Artifacts & Update Battery Saver LogicFixing modules turbo/artifacts

ariefnr:
Hello ALEAPP Team,

I hope you are doing well.

I encountered some issues while analyzing the Turbo_Battery and Turbo_Bluetooth artifacts using the 2022 CTF - Android-001.tar image (from Digital Corpora). The execution resulted in crashes due to AttributeError (when using Tar seeker) and NameError (undefined variables).

I have refactored deviceHealthServices_Battery.py to fix these crashes and updated the SQL logic based on the forensic data found in the image.

Here is the summary of changes:

1. Fix: `AttributeError: 'FileSeekerTar' object has no attribute 'directory'`
  - Issue: The code previously tried to strip the path using `seeker.directory`, which does not exist when the seeker is processing a .tar file (FileSeekerTar).
  - Fix: Changed to use `os.path.basename(file_found)` to safely extract the filename regardless of the seeker type.
2. Fix: `NameError: name 'file_found' is not defined` (in Turbo_Bluetooth)

- Issue: The logic was missing a proper iteration loop, causing it to reference `file_found` before it was defined.
- Fix: Implemented a proper for file_found in files_found: loop to handle multiple database files correctly.

3. Fix: `NameError: name 'time_offset' is not defined`

- Issue: The variable time_offset was passed to the timestamp conversion function but was never initialized.
- Fix: Initialized `time_offset = 'UTC'`.

4. Update: Battery Saver Logic Interpretation

- Analysis: Based on the artifact analysis, the previous SQL CASE statement for battery_saver was inconsistent with the data.
- Update: Adjusted the logic to:

- `0` -> 'Enabled'
- `2` -> 'Disabled'
- Others -> Returned as is.

5. Improvement: Error Handling

- Added try-except blocks when opening the SQLite database (open_sqlite_db_readonly) to prevent the entire artifact parsing from stopping if one DB file is corrupted/locked.

Traceback reference
```
AttributeError: 'FileSeekerTar' object has no attribute 'directory'
NameError: name 'file_found' is not defined. Did you mean: 'files_found'?
```

Here's what the error looks like before and bug fixing after
Before
Before Turbo Battery   Turbo Bluetooth 1

After
After Turbo Battery   Turbo Bluetooth

I have tested these changes against the CTF image, and the artifacts are now parsing correctly without errors.
Thank you for your hard work on this tool!

Best regards.

### DIFF
--- a/scripts/artifacts/deviceHealthServices_Battery.py
+++ b/scripts/artifacts/deviceHealthServices_Battery.py
@@ -37,94 +37,149 @@ from packaging import version
 from scripts.artifact_report import ArtifactHtmlReport
 from scripts.ilapfuncs import artifact_processor, open_sqlite_db_readonly, convert_ts_human_to_utc, convert_utc_human_to_timezone
 
+#Ok tambahkan UTC utk timezone dan tambahkan battery saver untuk 0 enable, dan 2 disable
 @artifact_processor
 def Turbo_Battery(files_found, report_folder, seeker, wrap_text):
     source_file_turbo = ''
-    turbo_db = ''
     data_list = []
-        
+    time_offset = 'UTC'  
+
     for file_found in files_found:
         file_found = str(file_found)
-        if file_found.lower().endswith('turbo.db'):
-            turbo_db = str(file_found)
-            source_file_turbo = os.path.basename(file_found)
+        if not file_found.lower().endswith('turbo.db'):
+            continue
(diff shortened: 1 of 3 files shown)


abrignoni:

PR #636, opened 2025-12-09
Fix Turbo_Battery artifact time_offset error (+ error handling on 2 other artifact reader) 

mikungg:
Before the fix:
image

After the fix:
image

I assume the problem was because the time_offset is undefined
image

So we need to makes sure we read for the timezone (and set it to UTC if no timezone is found just in case)
image

Turns out the timezone is America/New_York, this concludes the reader works properly
image

The image used for the testing:
https://digitalcorpora.s3.amazonaws.com/corpora/scenarios/magnet/2020%20CTF%20-%20Android.tar


abrignoni:

PR #640, opened 2025-12-10
Fixbug aleapp notif offline

andrypur-hub:
This pull request fixes an issue in offline notification history parsing where
ParseFromString(raw_data) caused failure on some corrupted protobuf data.
The fix adds safe skip logic when exception occurs.


abrignoni:

PR #642, opened 2025-12-10
fix airGuard Error

SchoolBoyJoJo:
This pull request addresses nine errors across multiple ALEAPP artifacts. The fixes include correcting invalid paths, handling missing files, preventing directories from being processed as files, and adding validation to ensure each plugin processes only the appropriate data. With these improvements, all previously failing artifacts now run successfully without errors.


abrignoni:
Thank you. This branch pulled in a large number of unrelated files during a fork sync so it cannot merge as is, and the airGuard error was since fixed. Closing.

PR #650, opened 2025-12-13
Fix validation appop setup wiz

hihanghoeng33:


abrignoni:
Thank you for working on these. The crashes were since fixed at the source, and the project avoids broad try/except wrapping because a swallowed error reads as no data. Closing this set.

PR #652, opened 2025-12-15
Add parsing for swissmeteo application

Th3Tul1p3:
Add parsing for swissmeteo application
- location of app opening and favourite places


stark4n6:
@Th3Tul1p3 please only submit the changes that are necessary for the PR, it looks like there are a lot of no update changes happening here. The best way to do this is start from a fresh fork of the current repository and add your additions.

PR #655, opened 2026-01-07
Add NetworkStats parser

0x4552494b:
This parser adds support for artifacts from the [Android NetworkStats API](https://developer.android.com/reference/android/app/usage/NetworkStats) located at `/data/system/netstats/`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ blackboxprotobuf
 fitdecode==0.10.0
 folium==0.14.0
 geopy==2.3.0
+mutf8
 packaging==20.1
 pillow
 polyline==2.0.0
--- a/scripts/artifacts/netstats.py
+++ b/scripts/artifacts/netstats.py
@@ -1,0 +1,258 @@
+__artifacts_v2__ = {
+    "netstats": {
+        "name": "Android NetworkStats",
+        "description": "Parses Android NetworkStats Data",
+        "author": "Erik (@0x4552494b)",
+        "creation_date": "2026-01-07",
+        "last_update_date": "2026-01-07",
+        "requirements": "mutf8",
+        "category": "Netstats",
+        "notes": "",
+        "paths": ("*/data/system/netstats/uid*"),
+        "output_types": ["html"],
+        "function": "parse_netstats"
+    }}
+
+# Tested against:
+# Google Pixel 3 - Android 11
+# Samsung Galaxy A32 (SM-A326B) - Android 13
+# Realme C11 2021 - Android 11
+
+
+import struct
+from enum import Enum
+
+import mutf8
+
+from scripts.artifact_report import ArtifactHtmlReport
+from scripts.ilapfuncs import logfunc, convert_unix_ts_to_utc
+
+
+# NetworkIdentity type labels
+# defined in https://cs.android.com/android/platform/superproject/main/+/main:frameworks/layoutlib/bridge/src/android/net/ConnectivityManager.java;l=118;drc=e29e1d58d803cb0f8f375d145294fcf78f6a8e57
+class NetworkIdentityType(Enum):
+    TYPE_NONE = -1
+    TYPE_MOBILE = 0
+    TYPE_WIFI = 1
+    TYPE_MOBILE_MMS = 2
+    TYPE_MOBILE_SUPL = 3
+    TYPE_MOBILE_DUN = 4
+    TYPE_MOBILE_HIPRI = 5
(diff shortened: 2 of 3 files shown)


0x4552494b:
Moved artifact icon into the artifacts_v2 header. The linter should pass now.

stark4n6:
@0x4552494b do you know of any public images that have test data? I haven't been able to find any just yet

0x4552494b:
Hi @stark4n6,

BinaryHicks [Android 11 Image](https://thebinaryhick.blog/2020/10/07/new-android-image-available-this-one-goes-to-11/) contains some data that can be parsed.

Unfortunately, I'm afraid the script can't comply with the new policy that all new artifacts must be LAVA ready. I can't think of any way to squeeze the data into a 2-dimensional view without omiting relevant info.

Here is a sample of the current output:

screenshot1

abrignoni:
Only idea that came to mind is to split it into two artifacts. One for the
network identity and another for the data buckets. The data buckets will
have a field for the network identity so it can be correlated to the
network identity entry in the previous artifact by the user.

Thoughts?

JamesHabben:
Agree with @abrignoni on the split. The identity artifact could even loop through the buckets to sum up things like total transfer value of each identity or count of transfers.

Edit: also it looks like you may be combining sessions for shared user id? If you split those into distinct records, this can allow sorting and filtering in lava, and the timestamps can get the timezone handling too.

0x4552494b:
Thank you guys for the suggestions. I currently see two problems with the splitting approach.

1. Network identities don't have a unique identifier. So to reference an identity from the "data buckets view" we'd have to introduce a new index that doesn't exist in the original data.

    _and more importantly_

2. In theory the file structure supports entries with multiple NetworkIdentities (the Android sources refer to it as [NetworkIdentitySet](https://cs.android.com/android/platform/superproject/main/+/main:packages/modules/Connectivity/framework-t/src/android/net/NetworkStatsCollection.java;drc=da6020a329e7f51c2b0b431bd14035ea24f805b7;l=520)). An entry could basically look like this:
```
.
├── NetworkIdentity
├── NetworkIdentity
├── Bucket
│   ├── Netstats
│   ├── Netstats
│   └── Netstats
├── Bucket
│   ├── Netstats
│   └── Netstats
└── ...
```

I didn't observe this during testing but we can't rule out the possibility that there will ever be an entry with more than one NetworkIdentity. In that case we'd need a way to link Buckets to multiple NetworkIdentities as well as convey that these NetworkIdentities belong together.

JamesHabben:
I think there is enough structure in the source data to support relational output.

The parser is already treating each parsed entry as a grouped structure containing:
- one or more NetworkIdentity records
- one or more bucket records

The source format may not expose a native unique ID for the identity set, but the grouping itself appears explicit from parse context.

For relational output, I think we could preserve that relationship by generating a synthetic grouping identifier during parsing. My preference would be to use the file offset of the entry structure rather than an arbitrary incrementing entry index.

For example:
- entry_offset
or
- identity_set_offset

Then both the identity table and bucket table could carry that value as the relational key.

I prefer offset over entry index because:
- it is derived directly from the source structure
- it provides forensic provenance/debug traceability
- it avoids introducing a purely parser-generated ordinal
- the meaning of the value is transparent to downstream consumers

An incrementing entry ID would also technically work, but offset feels cleaner and more defensible from a forensic/data lineage perspective.

There is still opportunity for aggregation/summarization as well. The offset/set identifier would simply preserve the original grouping relationship while still allowing aggregation by UID, identity fields, durations, totals, etc.

stark4n6:
@0x4552494b just fyi, tried on Josh's Android 11 image, got some errors for NoneTypes, so might have to add some extra file checks 
image

0x4552494b:
@stark4n6 Weird... I could have sworn this didn't happen when I originally wrote the script.

The problem is in [L264](https://github.com/abrignoni/ALEAPP/pull/655/changes#diff-1c9a2588faa1c780ad8db8bd847d2e18e8c491e2a11e5bf3048bc4045d4929f6R264), where I’m returning dummy data because otherwise the artifact icon wouldn’t be registered. Once the script is adapted to be LAVA-compatible and return actual data, the issue will resolve automatically.

abrignoni:
Thank you for this contribution and for being the first to take on the netstats store. A newer implementation of the same artifact just merged as #1052; it covers the versioned layouts without adding a dependency, so I am closing this one. Your work here is appreciated, and pull requests are always welcome.

PR #660, opened 2026-01-30
add DJI drone flight log parsers (TXT and DAT) for ALEAPP artifacts

riasramadan:
"This PR introduces two new artifact parsers for DJI drone flight logs:

1. DroneTXT.py: Parses detailed flight logs (.TXT) from DJI GO/Fly apps.

2. DroneDAT.py: Extracts data from DJI .DAT database files.

These modules are designed to enhance digital forensics capabilities regarding drone flight path reconstruction and telemetry analysis."

### DIFF
--- a/scripts/artifacts/DroneDAT.py
+++ b/scripts/artifacts/DroneDAT.py
@@ -1,0 +1,190 @@
+__artifacts_v2__ = {
+    "DroneFlightDATFiles": {
+        "name": "DroneFlightDATFiles",
+        "description": "Extracts cool data from database files",
+        "author": "@username",  # Replace with the actual author's username or name
+        "version": "0.1",  # Version number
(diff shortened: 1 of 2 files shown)


abrignoni:
Thank you for this one. The format work inside it is real and valued: the DAT record framing, checksum and XOR handling, and the TXT descrambling are the hard parts. What blocks it is order of operations: none of our test images carry actual DJI flight data, so there is no way to validate what a rework would produce, and a binary-format parser we cannot test against real logs is not something we can put in front of examiners. The right path is data first, parser second: once a real flight-log dataset is in our test corpus (the VTO Labs public drone datasets are the likely source), a rebuild validated against it from day one can land quickly, and the knowledge in this PR stays referenced here. Closing for now, with appreciation.

PR #677, opened 2026-03-19
Fix GroupMe not showing all chat messages

Underlyingglitch:
The GroupMe chat parser did not show all messages because it would ignore records that didn't have a `conversation_id` set (NULL value).

Changing the query to use a LEFT JOIN instead of a JOIN fixes this issue and shows all records in the table


stark4n6:
@Underlyingglitch merging, I know this parser is old so I'll probably have to update it to newer header and other things we've implemented in other parsers. It's on my list to work through each

PR #682, opened 2026-03-27
Add SBB Mobile parsing 

Th3Tul1p3:
Add parsing for purchased tickets, search history and places in SBB mobile train application


stark4n6:
@Th3Tul1p3 hi, do these DBs have -wal? if so we should apply the * to the file paths at the ends. Also if there are any timestamps in any of the parsers, if you move them to the front of the output they will get added to the timeline. Timestamps should also use the "datetime" formatting in headers too for LAVA output.

Let me know if you have questions.

Th3Tul1p3:
> @Th3Tul1p3 hi, do these DBs have -wal? if so we should apply the * to the file paths at the ends. Also if there are any timestamps in any of the parsers, if you move them to the front of the output they will get added to the timeline. Timestamps should also use the "datetime" formatting in headers too for LAVA output.
> 
> Let me know if you have questions.

Hi, I have fixed the * to accept wal files. for timestamp it's already the case. if you mean "datetime" when we are getting information with SQL, it's already the case. I not, can you be more precise please?

stark4n6:
@Th3Tul1p3 take a look here (line 137), if the header should be a timestamp add the "datetime" type handler https://github.com/abrignoni/iLEAPP/blob/main/scripts/artifacts/AMDSQLiteDB.py
image

PR #698, opened 2026-05-24
Added forensic modules for DeepSeek

RicardoBeny:
This PR adds initial ALEAPP support for DeepSeek Android artifacts through three new parsers:
- DeepSeek Chat Information
- DeepSeek User Information
- DeepSeek Chat Messages
The artifacts were rewritten to follow the current LAVA-compliant structure and updated to align with modern ALEAPP standards.

### DIFF
--- a/scripts/artifacts/Deepseek_ChatInfo.py
+++ b/scripts/artifacts/Deepseek_ChatInfo.py
@@ -1,0 +1,79 @@
+from datetime import datetime
+
+from scripts.ilapfuncs import (
+    artifact_processor,
+    open_sqlite_db_readonly
+)
+
+__artifacts_v2__ = {
+    "deepseek_chat_info": {
+        "name": "Deepseek Chat Info",
+        "description": "List of Deepseek chat sessions and last updated timestamps",
+        "author": "RicardoBentoSantos",
+        "creation_date": "2026-05-24",
+        "last_update_date": "2026-05-24",
+        "requirements": "none",
+        "category": "DeepSeek",
+        "notes": "",
+        "paths": ('*/data/data/com.deepseek.chat/databases/deepseek_chat_*.db'),
+        "output_types": ["html", "lava", "tsv"],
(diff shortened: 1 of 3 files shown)


abrignoni:
Please do not use utcfromtimestamp method since it is deprecated and produces naive times. :-)
Please use the time object definition in your data headers so LAVA can offset the timezone dynamically.
It will look like the attached image. What I do is I use the first item in the list to be the LAVA timestamp object in epoch unix time format and then do the second field in human readable format offset to UTC. That way i can offset the first timestamp in lava but still keep the utc one visible in the report no matter the first timestamp dynamic offset.

Screenshot 2026-05-24 at 11 00 48 PM

abrignoni:
Check out the end of the iLEAPP readme in the repo for a more detailed explanation.
The rest is looking pretty good!!!! Really, really good!!!
https://github.com/abrignoni/iLEAPP

RicardoBeny:
Updated both `DeepSeek_ChatMessages` and `DeepSeek_ChatInfo` to follow ALEAPP/LAVA timestamp handling requirements.

Changes made:

* Added LAVA-compatible datetime headers:
  `('Timestamp', 'datetime')`
* Added raw Unix epoch timestamps as the first column for dynamic timezone offset support in LAVA
* Preserved visible UTC timestamp columns for reporting consistency

JamesHabben:
For user info, we have a data type of `phonenumber` that use similar to the `datetime`.

For the message artifact, can you share what the source data looks like? While we have the HTML escape option, it doesn't work for LAVA output since it can allow for insecure code to run. If you share some samples we can help guide you to formatting it better.

RicardoBeny:
[DeepSeekData.zip](https://github.com/user-attachments/files/28230286/DeepSeekData.zip)

RicardoBeny:
Here is some data and some prints on how it looks in the ALEAPP report.

PR #713, opened 2026-06-02
update version to 3.7.0-dev.0

JamesHabben:


JamesHabben:
I dont have write/push permissions here so this is you @stark4n6

PR #716, opened 2026-06-08
feat: Adding Telegram parser

WriteBlocked:
This PR adds a Telegram android parser. It extracts artifacts from cache4.db and userconfing.xml, including: 

- Direct Messages
- Group Messages
- Channel Messages
- Account Preferences
- Contacts
- VoIP Calls
- Media Caches

### Note
This parser is internationally conservative. If I did not understand the value of an artifact, I did not include it. I tested this with 2 versions of Telegram that stored messages differently. If you have a version of Telegram that the messages were not read with, please provide me feedback and I will fix it.

### DIFF
--- a/scripts/artifacts/telegram.py
+++ b/scripts/artifacts/telegram.py
@@ -1,0 +1,2468 @@
+__artifacts_v2__ = {
+    "telegram_cache4_messages": {
+        "name": "Messages",
+        "description": "Extracts forensically relevant Telegram message records from cache4.db",
+        "author": "@WriteBlocked - Hiller Hoover",
+        "version": "0.1",
+        "date": "2026-04-13",
+        "requirements": "none",
+        "category": "Telegram",
+        "artifact_icon": "message-circle",
+        "notes": "",
+        "paths": (
+            '*/org.telegram.messenger/files/cache4.db*',
+            '*/org.telegram.messenger/files/account*/cache4.db*',
+        ),
+        "function": "get_telegram_cache4_messages"
+    },
+    "telegram_user_preferences": {
+        "name": "User Prefs",
+        "description": "Extracts Telegram account and user preference data from userconfig XML files",
+        "author": "@WriteBlocked - Hiller Hoover",
+        "version": "0.1",
+        "date": "2026-04-10",
+        "requirements": "none",
+        "category": "Telegram",
+        "artifact_icon": "settings",
+        "notes": "",
+        "paths": (
+            '*/org.telegram.messenger/shared_prefs/userconfing.xml',
+            '*/org.telegram.messenger/shared_prefs/userconfig*.xml',
+        ),
+        "function": "get_telegram_user_preferences"
+    }
+}
+
+"""
+Version history:
+
+4/6/2026: Cloned aLEAPP, set up pycharm for editing. started creating functions.
+4/7/2026: added userconfing.xml parsing.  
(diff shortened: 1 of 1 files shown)


abrignoni:
Thank you so much for the PR! The parser is super needed and useful. 
Here are a few needed changes:
1. LAVA support doesn't work.
2. Lots of pylint alerts.
3. Did a test run and group messages is empty. If empty it does not need to be in the report.
4. Remove any standalone test code when submitting.

Thank you so much again. This parser is key and the community will get a lot of value from it. Me included of course.

abrignoni:
Hi, and first off, thank you. I owe you an update and an apology for how long this sat.

Since June this PR has been waiting on the changes I asked for, and in the meantime Telegram support got built out in ALEAPP from another direction. That is on me for not circling back sooner, and I want to be straight with you about where things landed and make sure you get credit for what you contributed.

ALEAPP now has ten Telegram artifacts on main: messages, contacts, users, chats, accounts, peer details, chat details, VoIP call logs, auto-download settings, and save-to-gallery settings. So merging this PR as it stands is not going to work anymore, and I am not going to ask you to rebase 2,400 lines two months later. That would not be fair.

What I did instead was read your parser properly and take the parts where you were genuinely ahead of us. Two of them made it in, both credited to you in the code and the commit:

**Layer 132 peers.** You handle `TL_peerUser_layer132` and its siblings, which write the peer id as an Int32. We only handled the current 64-bit constructors, so a message stored by an older Telegram build raised on an unexpected constructor and every field after the peer was misaligned. That was a real bug in our parser, not a nice-to-have, and you caught it because you tested against two Telegram versions that store messages differently. That is exactly the kind of testing this stuff needs.

**VoIP call logs.** We were not touching `cache/voip_logs` at all. Your insight that the log file name is the call id is what makes it valuable, because that is the same id the phone call service message carries, so the log and the chat record tie together. Your handling of the timestamps is also right, and I checked it: the entries are local time with no zone, so using only their difference and taking the file modification time as the UTC anchor is the correct call. On my test image the last local entry and the mtime differ by exactly the device UTC offset, and the log span brackets the call duration the message reports. Two independent artifacts agreeing is the best kind of validation.

One thing I did not take: you define `0x20d51b14` as an alternate channel peer constructor. I could not find it in Telegram's generated Peer model, so I left it out rather than ship something I could not source. If you have a sample that produces it, I would genuinely like to see it, because that would mean I am missing something.

The one area where your parser is still ahead of ours is media object decoding, the photo and document trees. I did not adopt it, and the reason is corpus rather than quality: across the six Android extractions I have, almost all referenced media has been evicted from Telegram's cache, so the payoff looked small against the size of the work. If you have images where that is not true, I would rather have your implementation than write my own.

So, concretely, where I would love your help:

- If you want to open a focused PR for the media object decoding against current main, I will review it quickly this time. That is a real gap and it is your work.
- If you have a research image with group activity, saved drafts, or known deleted messages, that would unblock several readers that are currently only round-trip tested because nothing in my corpora exercises them.
- And if you spot anything in the current parser that is wrong, please say so. You already found one bug in it without trying.

I am going to close this PR, not because the work was not good, but because the ground it covers has moved and leaving it open pretends otherwise. Your name is in the source and in the commit history for both pieces that landed. Thank you for taking the time to write and share it, and I am sorry it took me this long to give you a proper answer.

abrignoni:
Closing as explained above. The two pieces adopted from this work are credited in the source and in #1031.

PR #720, opened 2026-06-16
Update compilation guide

Rugikex:
This PR fixes a compilation issue with newer scripts using 'zoneinfo'.
It also updates the 'Compile to executable' section in the README to match the clear, OS-separated instructions found in iLEAPP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ To create aleapp, run:
 pyinstaller scripts/pyinstaller/aleapp_Linux.spec
 ```
 
-To create aleappGUI.app, run:
+To create aleappGUI, run:
 
 ```
 pyinstaller scripts/pyinstaller/aleappGUI_Linux.spec
--- a/README.md
+++ b/README.md
@@ -27,16 +27,46 @@ To run on **Linux**, you will also need to install `tkinter` separately like so:
 
 To compile to an executable so you can run this on a system without python installed.
 
+_Windows OS_
+
 To create aleapp.exe, run:
 
 ```
-pyinstaller --onefile aleapp.spec
+pyinstaller scripts\pyinstaller\aleapp.spec
 ```
 
 To create aleappGUI.exe, run:
 
 ```
-pyinstaller --onefile --noconsole aleappGUI.spec
+pyinstaller scripts\pyinstaller\aleappGUI.spec
+```
+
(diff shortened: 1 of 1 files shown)


Rugikex:
Hi, I noticed the files have been updated since then. Is this PR still needed, or should I close it?

abrignoni:
Thanks for checking back @Rugikex, and sorry for the slow reply! 🙏 Here's where things stand — it's a split answer:

**The `zoneinfo` part is no longer needed.** It has since landed on `main` and is now present in all six spec files (including the `_Linux` and `_macOS` ones that didn't exist when you opened this). That's also why the PR is showing conflicts.

**The README part is still very much wanted** — arguably more than when you opened it. What's on `main` today is wrong: it tells people to run `pyinstaller --onefile aleapp.spec`, but there's no `aleapp.spec` at the repo root anymore (they all moved to `scripts/pyinstaller/`), and we've since added OS-specific specs the README says nothing about. So anyone following it hits a wall. Your OS-separated layout is exactly the right fix, and dropping `--onefile`/`--noconsole` is correct too since the specs already set those themselves.

If you're up for rebasing on current `main` and keeping just the README change, three small things to fix:

1. **Drop the leading separator** — `\scripts\pyinstaller\aleapp.spec` and `/scripts/pyinstaller/...` are absolute paths from the drive/filesystem root. They should be relative: `scripts\pyinstaller\aleapp.spec` / `scripts/pyinstaller/aleapp_macOS.spec`.
2. **Filename casing** — the actual files are `aleapp_macOS.spec` and `aleappGUI_macOS.spec` (capital `OS`). `aleapp_macos.spec` happens to work on macOS's case-insensitive default but breaks on case-sensitive setups.
3. **Add Linux** — we now ship `aleapp_Linux.spec` and `aleappGUI_Linux.spec`, so a Linux section alongside Windows and macOS would round it out.

No rush at all, and if you'd rather not bother with the rebase just say the word and we'll pick it up from here with credit to you. Either way, thanks for spotting the `zoneinfo` crash and for circling back on this. 🚀

Rugikex:
I have updated the README according to your feedback. Let me know if everything looks good now!
Just a heads up: Ileapp's README was updated in the meantime, but I kept the older version here.

PR #918, opened 2026-06-30
Update Knuddels parser and repository credentials

annkirpv:
Hi, 

I've updated my Knuddels parser based on new test data from a newer app version. The main change is that I adapted the parser to work properly with the LAVA viewer – I tested the integration and everything runs smoothly. I also updated the credentials since I renamed my repo. 

Could you take a look when you get a chance? Let me know if anything needs tweaking. 

Thanks!


JamesHabben:
The functions should change over to the current method of accepting the single `context` parameter rather than the legacy multi-parameter.

annkirpv:
> The functions should change over to the current method of accepting the single `context` parameter rather than the legacy multi-parameter.

Makes sense, thanks. I'll switch the functions to the single `context` parameter. I had tried that first but ran into a pylint issue with it... I'll work out the right pattern and push an update.

annkirpv:
> > The functions should change over to the current method of accepting the single `context` parameter rather than the legacy multi-parameter.
> 
> Makes sense, thanks. I'll switch the functions to the single `context` parameter. I had tried that first but ran into a pylint issue with it... I'll work out the right pattern and push an update.

Ah, now I see it! I'd only migrated one function to `context` and left the other two on the legacy parameters, which is what pylint was flagging. All three are on `context` now and lint passes. Thanks!

stark4n6:
@annkirpv we've gone away from version numbering in the headers, so I think those lines can be removed (in favor of the last updated timestamps that you already have)

JamesHabben:
Rotfl @ `rot13("password")`

annkirpv:
> Rotfl @ `rot13("password")`

Yeah, that's apparently how Knuddels protects user data 😄

PR #936, opened 2026-07-07
Update to Database name for MeWe app.

charpy4n6:
App name changed from app_database to app_v3.db. This is the only change made to this file. Tested with Hickman android 14 and also confirmed in Android OS 16 test image.


JamesHabben:
I think your branch is based on some really old main. The code diff is showing old code and reflecting a major change, and it's showing as conflicted and can't be merged.

charpy4n6:
I think I know what I did. I'll fix it today!

PR #955, opened 2026-07-14
Honeyboard clipboard artifacts - substantial rewrite of original

Al3x101:
I've extended the Honeyboard clipboard artifact significantly — WAL recovery, SEFT decoding, dual-schema support

Tested against: 
Android 16 - One UI 8.5
Android 14
Android 12

Parses:
1. ClipItem.db — clipboard text history live entries 
2. ClipItem.db-wal - WAL-recovered deleted entries 
3. clip files — clipboard screenshot images with Samsung SEFT trailer source-app extraction  

Key capabilities:
  - Dual schema: auto-detects old (caller_app_uid) vs new (caller_package_name)
  - packages.xml UID resolution: maps old-schema numeric UIDs to package names using the device's own installed-app list when available in the extraction
  - WAL deleted-entry recovery: self-contained SQLite WAL frame B-tree parser; last-frame-wins deduplication correctly implements SQLite read semantics
  - ClipData type decoding: 1=Text, 2=Intent, 3=URI, 4=HTML/Rich Text
  - Android user_id: 0=primary user, 150=Samsung Secure Folder
  - html and uri columns surfaced when present in the schema
  - Screenshot copy timestamp from parent directory name (ms epoch)
  - Screenshot EXIF: DateTimeOriginal / SubsecTimeOriginal / OffsetTimeOriginal
  - Samsung SEFT trailer decoded to source app package name (pure Python)
  - Image dimensions
  - Embedded thumbnail via check_in_embedded_media when Pillow available

### DIFF
--- a/scripts/artifacts/samsung_honeyboard_clipboard.py
+++ b/scripts/artifacts/samsung_honeyboard_clipboard.py
@@ -1,0 +1,810 @@
+"""
+Samsung Honeyboard — Clipboard History & Screenshot Clips (v1.0.0 - 2026-07-13)
+ALEAPP module
+
+Parses:
+  1. ClipItem.db  — clipboard text history, live entries and WAL-recovered
+                    deleted entries
+  2. clip files   — clipboard screenshot images with Samsung SEFT trailer
+                    source-app extraction
+
+Author:             Al3x101 (MSAB)
+
+Key capabilities:
+  - Dual schema: auto-detects old (caller_app_uid) vs new (caller_package_name)
+  - packages.xml UID resolution: maps old-schema numeric UIDs to package names
+    using the device's own installed-app list when available in the extraction
+  - WAL deleted-entry recovery: self-contained SQLite WAL frame B-tree parser;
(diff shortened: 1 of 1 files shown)


abrignoni:
Thanks so much for this, @Al3x101 — the WAL deleted-entry recovery, SEFT source-app decoding, and dual-schema handling are excellent additions, and the forensic approach is sound.

I've landed your work as #956, reworked to current ALEAPP conventions with full credit to you (and to @segumarc, whose original Honeyboard module this supersedes). The rework was needed because the module was built against an older ALEAPP: on current `main` it duplicated an existing `get_honeyboard_screenshot` plugin key (crashing PluginLoader at startup) and called `check_in_embedded_media()` with an old 6-arg signature (so the screenshot artifact silently produced 0 rows). #956 fixes those, moves the three artifacts to context form, co-extracts the WAL for the live table, collapses bind-mount mirror duplicates, and lints clean — validated on 4 Samsung images (Android 13–15).

Closing in favor of #956. Really appreciate the contribution — please take a look at #956 and flag anything you'd change.

PR #956, opened 2026-07-14
Samsung Honeyboard clipboard: WAL recovery + SEFT + dual-schema (rework of #955)

abrignoni:
Reworks @Al3x101's Honeyboard clipboard rewrite (#955) to current ALEAPP conventions and **supersedes** the original `SamsungHoneyboard.py` (@segumarc), which it replaces. Credit preserved for both in the artifact `author` fields and commit trailers.

## What this adds over the original @segumarc module
- **WAL deleted-entry recovery** — a self-contained SQLite WAL frame B-tree parser (last-frame-wins dedup); no external tools.
- **Samsung SEFT trailer decoding** — source app package from the proprietary trailer after the JPEG EOI (pure Python).
- **Dual-schema support** — auto-detects old `caller_app_uid` vs new `caller_package_name`; resolves old-schema UIDs via packages.xml when it's text XML.

## Why a rework of #955 was needed (verified against current `main`)
- 🔴 #955 added a `get_honeyboard_screenshot` plugin key that **already existed** in `SamsungHoneyboard.py` → `PluginLoader` raised `KeyError` at startup → **no plugins loaded at all**. Fixed by deleting the superseded module.
- 🔴 It called `check_in_embedded_media()` with a stale 6-arg signature → the call threw, was caught, and every screenshot row was skipped → **screenshot artifact produced 0 rows**. Fixed to the current `(source_file, data, name)` signature.
- 🟠 Legacy signature → **context form**; 9× broad-except / unused-arg / protected-access lint → **pylint 10.00**.
- 🟠 Live path narrowed to `ClipItem.db` dropped the WAL co-extraction; clip data lives in an un-checkpointed WAL (main db ~4 KB) → **fixed to `ClipItem.db*`**.
- 🟡 Rows dropped when a thumbnail couldn't be made (lost SEFT/timestamps) → **keep the row**; absolute per-row path → **relative**; naive/aware `sort` `TypeError` → **tz-aware**; noisy ABX packages.xml → **quieted**.
- **Mirror dedup:** collapses the same clip parsed from `data/data`, `data/user/0`, and `data_mirror` bind mounts (was inflating counts, e.g. one clip shown 3×).

## Validation — 4 Samsung images, 0 errors each

| Image | Android | Live | Deleted | Screenshot |
|---|---|---|---|---|
| SamsungS20 | 13 | 11 | 0 | 1 |
| SamsungA53 | 14 | 1 | 0 | 1 |
| Sharon | 14 | 19 | 0 | 17 |
| Anne | 15 | 8 | 0 | 3 |

Live/screenshot counts reproduce @segumarc's original sample_data (SamsungA53 shows 1 after mirror-dedup — it genuinely has a single clip). Verified end-to-end: the WAL engine decodes real records (`frame 8/page 4 → "069214"`); SEFT resolves the source app (`com.android.chrome`); EXIF capture time (with subsec/offset), dimensions, and thumbnails populate; dual-schema detects `Old (UID)`. `deleted=0` is correct — these WALs hold only live entries.

Closes #955 (superseded).


Al3x101:
Thanks' Alex, nice update, works well with my data sets. 
I should really keep a better eye on the LEAPP version so I don't mess up again.

PR #958, opened 2026-07-18
Add Smart Sidebar File Dock artifacts (OxygenOS/ColorOS transferdock)

akhil-dara:
New module scripts/artifacts/smartSidebarFileDock.py with two artifacts for the Smart Sidebar File Dock feature on OnePlus/Oppo/Realme devices (com.coloros.smartsidebar). Items reach the dock by drag and drop or via the share sheet's Save to File Dock.

Smart Sidebar File Dock: parses the hidden cache on shared storage at Download/.com_coloros_smartsidebar/transferdock/. Every dock item is cached there as a full copy inside a Base64-named folder that decodes to <localFileId>|<groupUUId>|<epoch ms added>|<name tail>. Media dragged out of apps is saved as
OPLUSDRAG_<source package>_<view>_<local time> files, preserving content from ephemeral apps (e.g. Snapchat) with source app attribution; text selections and links are stored as transferdock_<ms>.txt and link_file_<ms>.link files; other files keep their original names. Items persist for years and are hidden from MediaStore by a .nomedia file.

Smart Sidebar File Dock Db: parses the item registry in the private Room database (data/com.coloros.smartsidebar/databases/smartSidebar.db), including original source content:// URIs, MIME types, sizes, MD5s, timestamps and the transfer_dock_deleted_file_info table of removed items. The db* search pattern co-extracts the WAL because the main db is typically empty with schema and rows resident in the un-checkpointed WAL. Rows join to the cached copies via Local File ID / Group UUID.

Validated on two physical devices (282-item and 4-item docks, zip and tar inputs, including a 9.7 GB full file system image) and synthetic test data covering all item types and malformed inputs.


abrignoni:
Excellent work on this one, the documentation quality is a model for new artifacts. Two small fixups pushed to your branch: the ('File Size', 'bytes') header declared a type nothing implements yet, and an unrecognized type still travels into the LAVA manifest where it would activate the day a renderer exists, so it is plain text now. And the ephemeral-apps line now says the cached copy persists independently of the source app, labeled as observed on your devices, since the cache cannot show what the source app did. Verified with a synthetic round-trip of all three item types from your documented folder format (decode, timestamps, source app attribution all correct) and a clean run across our 12 test images, none of which carry ColorOS. A dock-bearing image for the corpus would be welcome. Merging. Good stuff.

PR #969, opened 2026-07-20
Add Mister Skinnylegs

AlexC-CCL:
This PR integrates our tool Mister Skinnylegs into ALEAPP to provide parsing of website artifacts from Chromium based browsers and browser components in extractions (Mozilla support can also be added).

Skinnylegs, like the LEAPPS  is a plugin-based architecture so I have built a sub-plugin loader which dynamically builds LEAPP artifacts at runtime, so ALEAPP should only need to update the version of Skinnylegs it is targeting to automatically get those plugins. There have been quite a lot of changes in Skinnylegs and our upstream browser modules in order to make things LEAPPable, but they forced my hand to make positive changes anyway!

In order to achieve dynamic loading, I did have to make a small, non-breaking change to artifact_processor to allow LEAPP plugins working in this way to inject globals, module and function name, as it was only seeing the lambda and none of the module attributes. (It would, theortically I think, be possible to achieve this by AST but I think that's likely to be a little arcane for a lot of contributors' tastes...).

I have added a new dependency for mister_skinnylegs, which I've pinned to a specific release. This implicitly also introduces upstream dependencies for some of our libraries, which does duplicate some of the modules in the "scripts/ccl" (everything but FCM, I think). I don't believe this will cause any collisions because of how they are currently being imported, but in future, assuming that there aren't massive changes to the LEAPP version of those modules, it might make sense to consolidate those.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ polyline==2.0.0
 protobuf==3.10.0
 PyCryptodome
 pytz
-mister_skinnylegs @ git+https://github.com/cclgroupltd/mister-skinnylegs.git@v0.0.16
+mister_skinnylegs @ git+https://github.com/cclgroupltd/mister-skinnylegs.git@v0.0.17
 simplekml
 wheel
 xlsxwriter==3.1.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ polyline==2.0.0
 protobuf==3.10.0
 PyCryptodome
 pytz
-mister_skinnylegs @ git+https://github.com/cclgroupltd/mister-skinnylegs.git@v0.0.17
+mister_skinnylegs @ git+https://github.com/cclgroupltd/mister-skinnylegs.git@v0.0.18
 simplekml
 wheel
 xlsxwriter==3.1.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,7 @@ polyline==2.0.0
 protobuf==3.10.0
 PyCryptodome
 pytz
+mister_skinnylegs @ git+https://github.com/cclgroupltd/mister-skinnylegs.git@v0.0.16
 simplekml
 wheel
 xlsxwriter==3.1.1
--- a/scripts/artifacts/browserArtifactsViaMisterSkinnylegs.py
+++ b/scripts/artifacts/browserArtifactsViaMisterSkinnylegs.py
@@ -66,7 +66,7 @@ class LeappArtifactStorageBinaryStream(ArtifactStorageBinaryStream):
         return self._stream.write(data)
 
     def close(self) -> None:
-        self._reference = check_in_embedded_media(self.source_file, self._stream.raw)
+        self._reference = check_in_embedded_media(self.source_file, self._stream.getvalue())
         self._stream.close()
 
     def get_file_location_reference(self) -> str:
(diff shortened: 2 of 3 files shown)


AlexC-CCL:
Looking at those checks, I'll see what it will take to target 3.10...

AlexC-CCL:
Hopefully the dependencies should install now!

AlexC-CCL:
OK, so at this point:

- Python Lint Check is failing on stuff in ilapfuncs which I haven't actually touched
- 3.10 runtime contract is failing on some stuff in NQ_Vault(?) which is nothing to do with me and on IntEnum not supporting auto() in python 3.10 in a dependency which isn't used in this implementation
- 3.11 runtime contact is failing on NQ_Vault

So, we are only failing on 3.10 due to a missing feature in a module which isn't currently called.

stark4n6:
@AlexC-CCL I think the NQ stuff is just admin tests so should be okay, @Johann-PLW can you give a quick peek at the ilap_funcs code changes?

AlexC-CCL:
As discussed - reverted changes in `artifact_processor` and switched to hardcoded, concrete artifacts

AlexC-CCL:
Failing tests are as before: either not my code or the IntEnum auto feature in py 3.10.

abrignoni:
Hey Alex,

Thanks for this PR and for the changes along the way. The new approach after the revert is much cleaner. I like it a lot.

Before merging I tested with real data: 4 full Android extractions from our test corpus (around 110 GB total: Samsung S10, S20, A53, and a Xiaomi Poco X7). It worked great. Installed clean on Python 3.14, all 28 artifacts loaded and ran, no crashes, and we got real records in the HTML, TSV, and LAVA outputs on every image (history, searches, localstorage, sessionstorage). The Bing and Google search results alone make this worth having. Nice work.

We merged it, and to keep things moving I made some small adjustments on our end right after (see the follow up commits on main). Quick rundown so future updates can stay in sync:

1. **Guarded the mister_skinnylegs imports.** If the package is not installed, only the browser artifacts get skipped now instead of ALEAPP failing to load all artifacts.
2. **Media fields in process().** The header was added as `("Media", k)` but ALEAPP expects `(column_name, 'media')`, and the row lookup would KeyError on the tuple key. Nothing in v0.0.18 uses media fields yet, but it would break the moment one does.
3. **app_xwalk profiles.** WeChat has a Crosswalk webview (`app_xwalk_*`) that was not recognized, so the cache step got skipped and the warning printed once per artifact. Now treated as a webview, and the warning only prints once per run.
4. **Source paths.** process() added the profile folder once per record instead of once per folder. On one image that made a source string of 789K characters for 20 folders. Now deduped.
5. **Formatting.** Matched the `__artifacts_v2__` block style to the other modules in the repo, since our maintenance and doc scripts run across all of them.

Two things we could use your help with on the skinnylegs side:

- Could you pin ccl_chromium_reader and ccl_mozilla_reader in mister_skinnylegs and cut a new tag? They currently install from master with no version, so a fresh ALEAPP install can change under our pin. We tried pinning them in our requirements.txt and pip refuses (conflicting URLs with the ones in your package metadata), so a lock file with --no-deps was our workaround. Pinning upstream is the real fix.
- Heads up that pip is pulling zstd 1.5.7.3, which was yanked from PyPI for not being thread safe.

If you base future updates to the ALEAPP module on the version now on main, everything should stay in sync. Thanks again for the work on this and for mister_skinnylegs itself. Great addition for anyone doing Android browser forensics.

abrignoni:
One last thing, can the artifacts be under a category named "MSL Web Artifacts"?
That way they are easy to find and provenance.

PR #1019, opened 2026-08-03
Feature: run each plugin in its own subprocess with skip/abort on Ctrl+C

Abu-Huraira21:
Summary

  Adds a --mp_per_plugin / --mp flag that runs each plugin in an isolated subprocess instead of
  in-process. Solves a common pain point during triage: a single slow/hung plugin currently forces you to
  kill the whole run (losing all progress) since there's no way to skip past it.

  Behavior

  - First Ctrl+C during a plugin: terminates that plugin's subprocess only, run continues to the next
  plugin.
  - Second Ctrl+C within 5s: aborts the run after the current plugin.
  - SIGUSR1/SIGUSR2 work as alternate skip triggers (useful when SIGINT is awkward to send, e.g. under
  sudo).
  - Debounces duplicate SIGINT delivery from sudo, which sends SIGINT to the whole process group and can
  otherwise double-fire the handler.
  - Default behavior (no flag) is unchanged — same in-process execution as today.

  How it works

  - Each subprocess reconstructs a minimal PluginLoader, a lightweight SeekerProxy (parent has already
  resolved files_found, so no re-searching needed), and re-opens the existing LAVA sqlite DB.
  - Only picklable deltas cross the multiprocessing.Queue back to the parent: icons_delta,
  lava_artifacts_delta, and meta_modules_delta (per-module metadata for _lava_data.lava). No plugin state
  itself is shared.
  - Parent merges these deltas after each subprocess returns.

  Testing

  Verified against a full local extraction run (625 plugins, --mp_per_plugin) — output (_lava_data.lava,
  HTML/TSV reports, LAVA sqlite) matches an in-process run byte-for-byte aside from ordering-independent
  fields. Also exercised the skip/abort paths manually via Ctrl+C mid-run.

### DIFF
--- a/aleapp.py
+++ b/aleapp.py
@@ -545,6 +545,13 @@ def crunch_artifacts(
                         icons.setdefault(cat, {}).update(arts)
                     for cat, arts in result.get('lava_artifacts_delta', {}).items():
                         _lavafuncs.lava_data['artifacts'].setdefault(cat, []).extend(arts)
+                    for module_info in result.get('meta_modules_delta', []):
+                        existing = next((m for m in _lavafuncs.lava_data['meta']['modules']
+                                          if m['module_name'] == module_info['module_name']), None)
+                        if existing:
+                            existing['artifacts'].extend(module_info['artifacts'])
+                        else:
+                            _lavafuncs.lava_data['meta']['modules'].append(module_info)
                 else:
                     logfunc('Error in {} [{}]: {}'.format(plugin.name, plugin.module_name, result.get('error')))
                     if result.get('traceback'):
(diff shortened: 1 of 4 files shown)


abrignoni:
Thanks for this, @Abu-Huraira21. This is the ALEAPP half of iLEAPP #1829, and it is being closed for the same reasons, written up in full there.

In short: the case that opened iLEAPP #1227 did not reproduce on the reporter's own backup (iLEAPP #1226), and what shipped instead is a start time and elapsed seconds for every artifact in all five LEAPPs (#1221 here). The subprocess model also leaks run-level state that the parent needs for the report: `identifiers` set by the four modules that call `device_info()` never returns from the child, so the Device Details page comes out empty; a `--custom_artifacts_path` run fails in the child, which builds the default loader; `seeker.search()` from a child returns nothing on zip or tar input; and a skipped artifact is not recorded anywhere an examiner reads.

The isolation is worth having, for crash containment and per-artifact memory as much as for skip, and it will be rebuilt as one design across the five cores. Your two PRs are the reference for that work.

PR #656, opened 2026-01-07
Add WhatsApp log files artifact

mpolastro:
## Description
This artifact parses WhatsApp internal event log files, extracting relevant metadata and activity records that help understand application behavior and user interaction over time, supporting forensic timeline analysis.

In addition, when the WhatsApp database (`wa.db`) is available, the script attempts to correlate log entries with contact information by matching phone number suffixes, suggesting the most probable contact related to a given action.

## Targeted files
- `*/com.whatsapp/files/Logs/whatsapp-YYYY-MM-DD.1.log`
- `*/com.whatsapp/files/Logs/whatsapp-YYYY-MM-DD.1.log.gz`
- `*/com.whatsapp/databases/wa.db`

## Extracted data
- Event types (e.g., message sent, received, deleted, conversation window focus)
- Associated timestamps
- Log line context and source file name
- Suggested related contact (when `wa.db` is available)

## Notes
- Follows the ALEAPP artifact structure
- Uses best-effort database schema discovery to support different WhatsApp versions
- Tested using local sample logs and databases

Suggestions and feedback are welcome.

Note: This PR supersedes a previous PR targeting an older ALEAPP version.

PR #1033, opened 2026-08-05
added WhatsAppLog indexer

LvRijn:
The added artifact indexes the WhatsApp.log files found at /data/data/com.whatsapp/files/Logs/. These log files contain useful information like WiFi connects/disconnects and the difference between device clock and NTP time.

Historic log files are gzipped in the same folder. The data from these files is also added with this script.

Examples:
ntp

screenshot


stark4n6:
@LvRijn I see you're spitting out the complete log to the column, do you think that the log info can be broken out further? I see timestamps and some other possible important info

LvRijn:
Its possible to split a log line into log_level, log_type, log_worker and log_content. The contents of the logs are too varied and not enough are understood, for now, to effectively parse them. In further iterations of this artifact we could add a column that would categorize certain logs. 

The reason I suggested the artifact this way is that the interface makes it simple to filter the results. Log level is in my personal use case not that interesting of a distinction. If you wish that this distinction is added, I will gladly do so. 

I've added an WhatsApp.log example file for your convenience. 

[whatsapp.log](https://github.com/user-attachments/files/30779613/whatsapp.log)

abrignoni:
Thank you for this and for posting the example log, which directly improved the merged artifact: the tz header your sample surfaced is now how the WhatsApp log timestamps get their UTC conversion in #656. That PR's token-based extraction covers the event reporting, so I am closing this one rather than indexing every line, which at these volumes would swamp the reports. The raw lines stay one click away in the source file. Your help is appreciated.

PR #1035, opened 2026-08-06
Report the file when a database will not open

abrignoni:
## The problem

`open_sqlite_db_readonly()` returns `None` when it cannot open a database. I audited every call site: **257 of 268 use the handle without checking it.**

Running the SMS/MMS modules against a copy of `mmssms.db` whose permissions blocked reading gave:

```
File "scripts/artifacts/smsmms.py", line 137, in _rows
  db.row_factory = sqlite3.Row
AttributeError: 'NoneType' object has no attribute 'row_factory'
```

with the same failure from `does_column_exist_in_db()`. The log line right above it already said `Error with <path>: - unable to open database file`, so the real cause was known and simply not carried forward.

## The fix

Adding a check at 257 call sites would be a huge diff, and each one needs its own "empty result" shape — often the headers are not even defined yet at that point. So the fix goes where the `None` comes from.

`open_sqlite_db_readonly()` now raises `SQLiteDatabaseError` naming the file, and callers no longer have to check the handle. `crunch_artifacts()` logs it and skips the artifact like any other error, minus the traceback that used to bury the message:

```
Reading SMS/MMS artifact had errors!
Error was Could not open database read-only: /path/to/mmssms.db
```

That covers all 257 sites at once, and a new module written in the usual style is correct without anyone remembering to add a check.

Helpers that probe a database whose absence is a real answer keep the old behaviour through the new `open_sqlite_db_readonly_or_none()`: `get_sqlite_db_records`, `does_table_exist_in_db`, `does_view_exist_in_db`, and the eight artifact call sites that already checked the handle (Deepseek x3, Grok, vaulty_files, waze). Those are unchanged in behaviour.

`does_column_exist_in_db()` was the one helper that did not check, and it is the call in the reported crash. It now reports "no" like its siblings. Its query string also moved above the `try`, so a failure on the first statement no longer trips a `NameError` in the `except`.

## Checking it

Reproduced the original failure against the fix — the `AttributeError` is gone and the error names the file. A readable database is unaffected.

New `admin/test/scripts/test_unreadable_database.py` pins both halves: the probe helpers stay tolerant, everything else fails loudly enough to name the file. It skips itself where the filesystem does not enforce the read permission (Windows, or running as root).

- 89 tests pass in `admin/test/scripts`
- `tests/core/test_sqlite_uri_paths.py` passes
- pylint `--disable=C,R` introduces no new warnings
- `check_claim_language.py` passes

## Notes for reviewers

The judgment call here is that this changes a public helper's contract rather than fixing 257 call sites, so the blast radius is wider than the 9-file diff suggests. Worth pushing back on if you disagree.

**What changed for whom:**

- `open_sqlite_db_readonly()` raises instead of returning `None`. Every artifact module calling it is affected. For the 257 sites that never checked the handle, this only changes the error they produce - they aborted the artifact before and they abort it now, so there is no behaviour regression, just a message that names the file.
- The 11 sites that did check are moved to `open_sqlite_db_readonly_or_none()` and are behaviour-identical.
- Out-of-tree artifact modules that rely on the `None` return would need the `_or_none` variant. I can only see this repo, so I cannot rule that out for anyone maintaining private modules.

**The alternative I did not take:** leave `open_sqlite_db_readonly()` alone, add a raising variant, and migrate the 257 call sites to it. Same end behaviour, no contract change, but a ~185-file diff and the original footgun stays in place for new modules. Happy to switch if the team prefers that trade.

**Known limitation:** a module that loops over several databases still aborts the whole artifact on the first unreadable file, rather than skipping that file and continuing. That is what happens today too, so it is not a regression, but it is not an improvement either. Per-file resilience needs per-module judgement and would be a separate change.

**On the red checks:** the first runs landed during the GitHub Actions major outage on 6 Aug (incident opened 15:22 UTC). Those jobs were cancelled at setup with no steps executed and no logs, so they are not real results. `runtime-contract (3.10)` did run and passed - that is the job that runs the whole `admin/test/scripts` suite and imports every artifact module. The rest will be re-run once Actions recovers.

### DIFF
--- a/admin/test/scripts/test_unreadable_database.py
+++ b/admin/test/scripts/test_unreadable_database.py
@@ -1,0 +1,108 @@
+"""Unreadable evidence database tests.
+
+open_sqlite_db_readonly used to return None when a database would not open,
+and almost every artifact used the handle without checking it. An evidence
+file that was locked, truncated or simply unreadable therefore surfaced as
+"AttributeError: 'NoneType' object has no attribute 'row_factory'" further
+down the module, naming neither the file nor the reason - even though the
+real cause had already been logged one line earlier.
+
+open_sqlite_db_readonly now raises SQLiteDatabaseError naming the file, and
+the runner in aleapp.py logs that and skips the artifact. Helpers that probe
+a database whose absence is a legitimate answer - does_table_exist_in_db and
+friends - keep the tolerant behaviour through
+open_sqlite_db_readonly_or_none.
+
+These tests pin both halves: the probe helpers stay tolerant, and everything
+else fails loudly enough to name the file.
+"""
+import os
+import sqlite3
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+ROOT_DIR = os.path.dirname(
+    os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+)
+if ROOT_DIR not in sys.path:
+    sys.path.insert(0, ROOT_DIR)
+
+from scripts.ilapfuncs import (SQLiteDatabaseError, does_column_exist_in_db,
+                               does_table_exist_in_db, does_view_exist_in_db,
+                               get_sqlite_db_records, open_sqlite_db_readonly,
+                               open_sqlite_db_readonly_or_none)
+
+
+class TestUnreadableDatabase(unittest.TestCase):
+    def setUp(self):
+        self.temp_dir = tempfile.TemporaryDirectory()
+        self.addCleanup(self.temp_dir.cleanup)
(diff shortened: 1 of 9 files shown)


abrignoni:
Closing this for now - will come back to it later.

The underlying issue is still there: open_sqlite_db_readonly() returns None and 257 of its 268 call sites use the handle without checking, so an unreadable database still surfaces as an AttributeError that names neither the file nor the reason.

Leaving it open would just age. Reopening this PR restores the branch and the work if we pick the same approach, and the discussion in the description covers the trade-off between changing the helper's contract and migrating the call sites.

PR #1038, opened 2026-08-06
Added Claude Artifacts

bcbtree:
parsers for claude conversations, messages, and projects, as well as the account info for the profile


stark4n6:
Hi @bcbtree, thanks for the contribution. Do you think these could be combined into one .py? They seem simple enough to do so.

bcbtree:
@stark4n6 updated to one .py

abrignoni:
Thank you for these, Brandon. Great addition. I pushed a small maintainer fixup to your branch before merging: renamed last_updated_date to last_update_date (the field the framework reads), removed the data_views Attachment key (LAVA has no such key, so it did nothing; the image file name still shows as a regular column), punctuated the notes, made the conversation join LEFT for completeness, and added sample_data from validation runs against two of our test images. Both images reproduce your parser's output exactly: 1 account, 14 conversations, 78 messages and 1 project on one, 1/1/8/0 on the other, matching the database contents row for row. Good stuff.

PR #1052, opened 2026-08-07
Added Netstats Plugin

AlexC-DF:
This plugin parses the contents of the system/netstats folder and provides network data usage stats for apps/system, processes, including what type of connection the usage took place on (e.g., WiFi, mobile network, VPN), and identifiers for that network (e.g., SSID and subscriber ID).

There are likely refinements that can be made to improve how the data is reported especially to reduce the volume of records (e.g., I think the "buckets" can be safely combined, I just need to work out how to approach that), but as a first pass, I think it already provides value.


abrignoni:
Excellent work, thank you. Ran it against our 12 test images: rows on every one, 1,894 to 62,999. Found one crash: an image with an encrypted packages.xml killed the whole artifact before any netstats file was read. I pushed a fixup to your branch: guarded that parse (bare UIDs when the map is unavailable, and that image now returns 6,794 rows), rejected the legacy v2-4 layouts instead of reading them with the v16 layout, relabeled UID -1 per AOSP UID_ALL, fixed the description, filled the notes, and added sample_data from the runs. Merging. Good stuff.

AlexC-CCL:
I would argue that the change to the labelling for UID value -1 could be misleading

See https://cs.android.com/android/platform/superproject/+/android-latest-release:packages/modules/Connectivity/framework-t/src/android/net/NetworkStats.java;drc=27189f4dff3c016d7a635b3cfc12fd02e53e238e;l=82 where it is described specifically in the comments as "value when UID details unavailable". The use of "All" in the output implies that all user ids are generating activity rather than just some subset (and maybe just 1) which hasn't been recorded.

PR #1140, opened 2026-08-16
Add Opera Browser tab/session-state artifacts for Android

Gear-I:
**Summary**

Adds opera_tabs and opera_tab_navigation, reading Opera's own session_db (the tab, recently_closed_tab, navigation_entry, and tab_order tables) rather than the shared Chromium History database.

**Important scoping note:** 
Opera's regular browsing history, search terms, bookmarks, autofill, cookies, and saved logins are already covered by this project's chrome.py family; those modules already glob for app_opera alongside Chrome, Samsung Browser, and WebView, since Opera is Chromium-based and stores that data in the identical schema. This was confirmed by running the real pipeline against a real device and watching Opera's own visits appear in the existing "Web History" / "Web Visits" artifacts (25 and 32 rows respectively on the validation device). This PR deliberately covers only what's unique to Opera and isn't already collected: its session/tab-state store, which persists independently of History and can recover a tab that was closed before its pages ever finished syncing into History's own tracking.

Opera Browser - Tabs: one row per tab, open or closed, with its tab-bar position, the page it was showing, its navigation-entry count, and whether a closed tab was later restored.
Opera Browser - Tab Navigation History: the full back/forward navigation stack inside each tab, in order, with the currently active page flagged. Opera's own opera-internal://search virtual-URL scheme is decoded back into the literal text typed into the address bar for a search; entries that aren't a search are left blank rather than guessed at.
What was deliberately left out

Two other Opera-only stores were inspected and are not included in this PR:

reading.db (the Reading List / "save for later" feature) had zero rows on the validation device, so its column mapping could not be confirmed against real content.
searchengines.db holds only the static list of built-in search engines — configuration, not a user action.

**How this was validated:**

Ran end to end through the real aleappGUI.py against a real device extraction (Pixel 7a, Android 14) with a documented public action sheet (Josh Hickman, thebinaryhick.blog) describing an exact browsing session: three normal tabs (mlb.com --> an MLB.com article, a "mobile phone forensics" search -->  cellebrite.com, and a "shelley duvall" search) plus three private tabs, with two tabs explicitly closed along the way.

Result: 3 tabs recovered (2 open, 1 correctly flagged as closed), 8 navigation entries, titles decoding cleanly (Opera stores them as UTF-16LE blobs), and both address-bar searches decoded to their exact typed text ("mobile phone forensics", "shelley duvall"). The real MLB.com article slug (ben-joyce-throws-104-5-mph-fastball) and the cellebrite.com URL match the action sheet's descriptions precisely. Zero private-tab data appears anywhere in session_db, consistent with the action sheet's private-tab actions and an empty databases-off-the-record folder on this device.

One genuine, documented gap surfaced during validation: the action sheet's "Ohtani reaches 200-homer milestone" tab click does not appear in that tab's navigation stack. This isn't a bug — standard browser back/forward semantics mean going back and then clicking a different link overwrites the forward entry that link replaced, so a page that really was viewed can still be absent from the tab's own stack. It remains fully recoverable from the generic Chromium History artifact, which this PR's notes field points reviewers to, so the two artifacts should be read together rather than either being treated as a complete visit log on its own.

Co-authored-by: Claude Version: Sonnet 5


abrignoni:
There is nothing wrong with using LLMs. I use LLMs but I also attribute them in the PRs and I expect every contributor to do so as well.

Gear-I:
Reword Opera artifact descriptions and simplify output_types

check_claim_language.py flags "every" as an unsupported completeness
claim in examiner-facing name/description fields (the check does not
read `notes`). Reworded both opera_tabs and opera_tab_navigation
descriptions to describe the record and the table it comes from
instead of asserting completeness - no behavior or data change.

Also switched output_types from the explicit
["html", "tsv", "timeline", "lava"] list to this project's
["standard"] shorthand, already used by 10 other artifacts. Per
check_output_types() in ilapfuncs.py, "standard" resolves to every
output type except kml, so this is functionally identical for these
two artifacts (neither emits location data).

Validated: python admin/scripts/check_claim_language.py reports 0
findings for OperaBrowser.py. pylint --disable=C,R: 10.00/10. Re-ran
aleapp.py end-to-end against the validated device's Opera data -
HTML and TSV both still generate under "standard", and row counts
are unchanged (3 tabs, 8 navigation entries).

Co-authored-by: Claude Version: Sonnet 5 (Also added in the actual parser script as "author": "@Gear-I, Claude"

abrignoni:

PR #1142, opened 2026-08-17
Add six Spotify artifacts support for Android

Gear-I:
# Add six Spotify artifacts support for Android 

## Summary

Adds six artifacts for `com.spotify.music`: `spotify_account`, `spotify_playlist_library`, `spotify_playback_activity`, `spotify_recently_played`, `spotify_now_playing_view`, and `spotify_artist_profile_views`. Spotify has no existing ALEAPP coverage. All formats are undocumented and were reverse-engineered directly against a real Pixel 7a/Android 14 extraction (Josh Hickman's public image), cross-checked against that image's published action sheet as the ground truth.

The first three artifacts came from the app's own databases and preferences files. The device's action sheet documents a considerably richer listening history (specific tracks across three separate sessions in January and July 2024) than those three alone account for, so the rest of the app's data directory was re-examined for other independent sources of evidence. `cache/http-cache` the app's HTTP disk cache turned out to still hold the most recent cached response to several of Spotify's own API calls, which is what the last three artifacts recover.

## What each artifact reports

**spotify_account** the signed-in account's canonical username, read from `spotify_preferences.xml`'s `crashlytics_user_id` (corroborated by the separate `event-sender-event-owner` value agreeing), the sign-in method (email/password vs. a linked provider), selected language, and the app's own first-launch time. The account's email address and display name are not stored anywhere on disk in this extraction — confirmed by searching the entire app data directory for both, so neither is reported.

**spotify_playlist_library** decodes `frecency.pb`, a per-account protobuf file under `files/settings/Users/<id>/` that Spotify uses to rank frequent/recent content. No public schema exists for this file; it was reverse-engineered directly against real data. Reports every playlist snapshot the file holds, with its timestamp and two raw usage counters. On the validated device, this recovered two playlists at the exact minute the action sheet documents them being added to the Library (2024-01-29, two entries a few minutes apart), plus a later snapshot roughly six months on showing one playlist's counters grown from 1/1 to 21/4, consistent with real repeated use rather than a static config value.

**spotify_playback_activity** decodes `event-sender.db`'s `Events` table, an internal analytics log whose `fragments` column is schema-less protobuf, decoded the same way this project already handles undocumented protobuf elsewhere (see `YouTube.py`'s `account.pb`/profile decoding). The table held 58 distinct event types on the validated device, the large majority pure app diagnostics with no forensic value (ad errors, cache reports, connection state, audio driver info). This artifact reports only the six event types confirmed to carry real content when decoded `Playback Segments`, `BoomboxPlaybackSession`, `CorePlaybackFinished`, `Download`, `AudioFileSelection`, `TrackNotPlayed` and only the fields directly confirmed against real data: literal `spotify:track:*` / `spotify:playlist:*` URIs, an opaque session-correlation ID Spotify's client reuses across a group of related events (normalized from either its raw 16-byte binary form or its hex-string form, so both encodings line up under one column), and short human-readable status strings. Purely numeric telemetry fields with no public schema (byte counts, bitrates, internal enum values) are left out rather than guessed at.

On the validated device extraction, this recovered a coherent, session-correlated 14-hour playback timeline for one real track (`spotify:track:1x5MjCffpcdHLf65eR9r3T`), ending in an Android Auto `TrackNotPlayed` event that references the same playlist URI recovered independently in `spotify_playlist_library`. Cross-checking event-sender.db's own row ID range showed why the action sheet's earlier sessions don't appear here: the table only ever holds a short rolling window (456 rows total on the validated device, none older than 2024-07-27 21:17 UTC) — it is not a historical log.

**spotify_recently_played** decodes a cached response from the `recently-played/v3` endpoint (schema-less protobuf, same decoding technique as `spotify_playback_activity`). Recovers a track URI, its playlist context URI, and a millisecond timestamp embedded in the response by Spotify's own server, not derived from any local file time. The endpoint is requested with `limit=50`, but the cache only ever holds the most recent response, so this reflects one snapshot, not a full history.

**spotify_now_playing_view** decodes cached `color-lyrics/v2` and `merch-npv-service/v1` responses, both requested by track ID for whatever is showing on the Now Playing screen. Reports the track ID (read directly from each cached request's URL), the first line of the time-synced lyrics recovered from the response body, and whether merch was also fetched for the same track. No track-ID-to-song-name table is hardcoded anywhere in this module — a lookup table built from one device's track IDs would not generalize to any other extraction — so the lyric line itself, verbatim, is what lets an examiner identify the song.

**spotify_artist_profile_views**  parses cached `artist-identity-view/v2` JSON responses directly for their own `name` and `artistUri` fields; no interpretation involved.

Both new track-level artifacts use the cached response file's own filesystem modification time as their timestamp column — the same convention this project already uses for browser cache artifacts (e.g. `browserCachechrome.py`) — documented in each artifact's `notes` as approximately when the content was cached/viewed, not a playback timestamp.

On the validated device extraction this recovered lyrics and artist views for four tracks/artists clustered within about 14 minutes of each other (20:31–20:45 UTC, 2024-07-27): Descendents, LMFAO, The Clash, and Vampire Weekend, in that order matching the action sheet's 16:31 (UTC-4) playback session, which names Descendents, LMFAO, The Clash, and Vampire Weekend in that same order. `spotify_recently_played`'s independently-recovered timestamp falls inside the earlier 12:38 to 12:56 session the action sheet documents for the same day.

## Validation

- Ran end-to-end via the real `aleapp.py` CLI against a scoped copy of the device's `com.spotify.music` data directory (all ~802 real artifacts, not a stub), copied with `cp -p` to preserve original file timestamps and confirm the output shows real 2024 times rather than copy-time artifacts.
- `spotify_account`: 1 row. `spotify_playlist_library`: 3 rows. `spotify_playback_activity`: 28 rows. `spotify_recently_played`: 1 row. `spotify_now_playing_view`: 4 rows. `spotify_artist_profile_views`: 4 rows.
- `pylint --disable=C, R`: 10.00/10.
- `admin/test/scripts/test_artifact_imports.py`: pass.
- `admin/scripts/check_claim_language.py`: 0 findings.

## Scope notes

- `auth-cache.db`'s `cachedClientInfo` table was checked and excluded — it is third-party OAuth client registration data, not the device user's own identity.
- `bmax-offline.db`'s `OfflineLicenseInfoEntity` and `exoplayer_internal.db`'s `ExoPlayerDownloads` were both checked and had 0 rows on the validated device despite the action sheet documenting two playlist downloads, so no offline-license/download-file artifact is included.
- `offline_lists.bnk` under the per-user settings folder does hold a real playlist URI in plaintext, but its container format is a proprietary Spotify serialization (not protobuf, not decodable with the tools this project already vendors), so it was left unparsed rather than guessed at.

 Co-authored-by: Claude Sonnet 5 <noreply@anthropic.com>


abrignoni:

abrignoni:
Thanks for this one. Before I add the CI fixture for it, one change request on the three http-cache artifacts.

Cache File Time comes from os.path.getmtime, which is a property of whatever copy of the cache the tool reads, so the value moves every time the evidence is re-staged and the artifact cannot be regression tested. The cache's own .0 entries already record when the request happened: OkHttp-Sent-Millis and OkHttp-Received-Millis, plus the server date header. On the Josh Hickman Android 14 image, all 8 responsive artist and lyrics entries carry OkHttp-Received-Millis, and they decode to 2024-07-27, the day before that image was acquired. Reading that value out of the .0 text gives a device-recorded time that does not change between copies, which is better evidence and lets the fixture gate in CI.

One smaller note, nothing needed from you: the three artifacts each glob the whole cache/http-cache/*, so a test fixture carries the same 354 files three times. Once the timestamp change lands I will cut the fixture trimmed to the responsive request and response pairs, which takes it from 17 MB to about 1 MB.

Happy to entertain a different approach if you have one in mind.

Gear-I:
Test data and fixtures are provided; trimmed the three http-cache zips down to just the matched request/response pairs (about 1MB total instead of the 17MB make_test_data.py produced by default), verified test_module.py still reports identical row counts before and after trimming.

abrignoni:
Thanks for making the timestamp change, and for the trim. I checked both end to end before touching anything: the four `Response Received Time` values decode to 16:31:09, 16:31:45, 16:36:19 and 16:44:49 device local, inside the documented 16:31 session to the second, the trimmed zips produce the same rows the untrimmed ones did, and the fixture is down to 1.06 MB.

Rather than send it back to you and wait, I pushed a commit to the branch with four fixes. Each one is measured against a constructed tree rather than reasoned about, and the numbers are here so you can check them.

**The Notes filter was dropping the one value the artifact promises.** `_summarize_message` skipped any decoded string containing a space, and `offlined file` is on every `AudioFileSelection` event. That string is what sits behind "whether a session was playing from a downloaded local copy rather than streaming" in the description, and the notes name it as an example. Removing the space test makes 7 of the 28 rows gain it and moves nothing else in the table. It matters more than a docs slip here: your own scope note says `bmax-offline.db` and `ExoPlayerDownloads` were both empty on this device, so those seven strings are the only local evidence left that anything was played from a download.

**Duplicate storage views were multiplying rows.** `_http_cache_pairs` keys on the full path, so `data/data/`, `data/user/0/` and `data_mirror/data_ce/null/0/` views of one cache entry become three separate pairs. On a tree carrying all three: artist views 4 to 12, now playing 4 to 12, recently played 1 to 3. `storagePathViews.unique_files()` is already on main and 95 modules import it. Wired in, those go back to 4, 4 and 1, and a single container image is unchanged.

**`_find_one` was dropping a second account.** The playlist library glob is `files/settings/Users/*/frecency.pb`, so that wildcard is the account id, and taking the first match stops at one. Two Spotify accounts signed in on one phone, one Android user, no exotic device needed: 3 rows reported where 6 exist, with no log line and no error. The same accessor dropped a second Android user's account row and playback rows. Every match is now read, and there is a new `Account Folder` column carrying the directory name as stored so two accounts' rows stay apart. On this device that name is the canonical username with `-user` on the end, so it joins straight to Spotify - Account.

**The located-at line named one file out of four.** `source_path` was reassigned inside the loop, so the report pointed the examiner at whichever cache entry happened to be matched last. It is newline joined now, which is what `artifact_processor` and the LAVA manifest expect.

Two smaller ones: `file_count` on the three trimmed zips still read 374, which was the pre-trim number, so it now records what each zip actually holds and the case description says the zips were trimmed; and a cached lyrics response that will not decode now logs instead of silently leaving the cell blank.

Snapshots re-recorded. The other four artifacts are byte identical row for row, and the two that changed changed only as described. The gate was red on exactly those two units before the re-record and is 31 PASS after. pylint 10.00, and the claim language, source path, report path and column order checkers are all clean.

One thing I left for you. The PR body says the recently played timestamp falls within that same session. It reads 12:54:19 local, about three and a half hours before the 16:31 session. That agreement was the support for reading that protobuf field as the server's play time, so it is worth another look rather than just editing the sentence.

abrignoni:
Correction to my last comment, and the error is mine, not yours.

I flagged the recently played timestamp because it reads 12:54 local while the PR body puts it in the 16:31 session. **The value is right and your reading of the field is right.** I checked it against every registered Android extraction instead of the one image, and it holds:

- On one extraction a single cached response carries **50 entries with 50 distinct times**, descending. So the field is per entry, not a collection stamp.
- On two others the same entries keep byte identical times across responses fetched weeks apart, five separate fetches on one of them. A value that does not move when the request moves is a property of the play event.
- Hickman's creation log for the Pixel 7a image documents a playback session from **12:38 to 12:56** local on 2024-07-27. The entry decodes to 12:54:19, inside it. The 16:31 session is a separate, later one, and it is when the app made the fetch, so the server returned what had been played before it.

What made it look wrong is worth recording, because the next person will hit it. The response's `mc-etag` header carries the newest entry's timestamp. That image cached exactly one entry, so the newest entry is the only entry and the two numbers are necessarily identical. On any response with more than one entry they differ. That is now in the artifact notes. The only thing actually wrong is the one sentence in the PR body.

## Twelve corpora

Since I was already there, I ran the module end to end against every registered Android extraction that holds Spotify data, twelve of eighteen, and recorded the results in `sample_data` including the zeros. Pushed as a second commit.

| corpus | account | playlist | playback | recently played | now playing | artist |
|---|---:|---:|---:|---:|---:|---:|
| galaxys10_a10 | 1 | 7 | 3 | 0 | 0 | 0 |
| cookbook_a11 | 1 | 0 | 0 | 0 | 0 | 0 |
| pixel3_a11 | 1 | 2 | 0 | 0 | 0 | 0 |
| pixel3_a12 | 1 | 10 | 3 | 0 | 0 | 0 |
| samsungs20_a13 | 1 | 2 | 1 | 6 | 2 | 0 |
| russell_pixel6a_a13 | 1 | 13 | 3 | 15 | 0 | 0 |
| sharon_a13 | 0 | 0 | 0 | 0 | 0 | 0 |
| pixel7a_a14 | 1 | 3 | 28 | 1 | 4 | 4 |
| russell_a14 | 1 | 14 | 0 | 10 | 1 | 1 |
| sharon_a14 | 0 | 0 | 0 | 0 | 0 | 0 |
| anne_a15 | 1 | 1 | 7 | 2 | 0 | 0 |
| kevin_pocox7_a15 | 1 | 21 | 2 | 50 | 2 | 0 |

Every run finished with its completion marker, exit 0, and a clean grep for the SQLite error vocabulary. The recently played counts were cross-checked against an independent decode taken straight from the archives, outside the tool: 84 entries either way, agreeing on all twelve.

Every zero above was confirmed rather than assumed. The two `sharon` images carry the preferences file with none of the account keys in it, which is an installed but never signed in app. `cookbook_a11`'s event database holds 4 rows and `russell_a14`'s holds 27, none of either being one of the six reported types.

## The Notes filter, on more than one device

The `offlined file` fix turns out to matter on four extractions, not one, and the field carries more than one value: **`offlined file`**, **`cached file`** and **`best matching bitrate`**. The first two are not the same thing, since one is a copy the user chose to download, and before the fix every one of those rows read just `audiophile`. That distinction is now stated in the notes.

## One boundary worth being straight about

On the registered corpora my earlier fixes changed no row counts at all. `pixel3_a11` and `pixel3_a12` do carry the container under two and three storage-view spellings, and `unique_files()` correctly collapses them, but the old `_find_one` happened to dedupe those too, and neither image has a responsive http-cache entry to multiply. No registered corpus has a second Android user or a second Spotify account. So the duplicate-view and second-account fixes are exercised only against trees I built for the purpose, and I would rather say that plainly than imply corpus proof they do not have.

Everything else is green: gate 31 PASS, pylint 10.00, and `validate_sample_data.py --registry` resolves all twelve keys with no findings.

abrignoni:
@Gear-I one leftover, and it is only the PR text. The code and the artifact notes are right, and this is merged, so nothing an examiner reads is affected.

Last line of the Now Playing View paragraph:

> `spotify_recently_played`'s independently-recovered timestamp falls within that same session.

It falls in the other one. The action sheet documents two playback sessions that day, 12:38 to 12:56 and 16:31 to 16:48, and the recovered value decodes to 12:54:19 local, inside the first. The 16:31 session is when the app made the recently-played fetch, so the server returned what had been played before it. That is why the two look adjacent without being the same session, and it is also the reason the value is worth having: it is independent of the session your other three artifacts captured, rather than a fourth confirmation of it.

If you want a replacement:

> `spotify_recently_played`'s independently-recovered timestamp falls inside the earlier 12:38 to 12:56 session the action sheet documents for the same day.

Happy to make that edit for you rather than have you go back into a merged PR. Just say.

To close the loop on my end: I raised that timestamp as a possible problem with your field mapping and I was wrong about it. Your reading is correct. What threw me is that on the Pixel 7a image the value is identical to the response's `mc-etag` header, and I read that as a collection stamp wearing a play time's label. That image caches exactly one entry, and the etag carries the newest entry's timestamp, so with one entry the two are necessarily the same number. On any response with more than one they differ, and one of the other registered extractions caches fifty entries with fifty distinct times. Good module, and thanks for turning the mtime change around quickly.

abrignoni:
Went ahead and made that edit, @Gear-I. One sentence in the Now Playing View paragraph, nothing else in the body touched. The original wording is quoted in my comment just above if you want it put back.

Gear-I:
Thank you @abrignoni.  :)

PR #1144, opened 2026-08-17
Add Threema account, contacts, and messages artifacts

Gear-I:
**Summary**

Threema had no existing ALEAPP coverage. This adds three artifacts, built and validated against a real Pixel 7a/Android 14 extraction (Josh Hickman's public forensic training image), and cross-checked against that image's published action sheet.

**Artifacts**

threema_account - Threema ID, nickname, and any linked mobile number, read from shared_prefs/ch.threema.app_preferences.xml. Threema has no email/password account, so the ID itself is the identity.

threema_contacts - Threema ID, display name, verification level, date added, and two flags, read from the decrypted contacts table in threema4.db. Verification level uses Threema's own three-state indicator (Unverified / Server-verified / Fully verified), corroborated on this device by a documented QR-scan verification event for the one contact recorded at the fully-verified level. "Group-Only / Removed" reports Threema's own AcquaintanceLevel field; per Threema's own Android source, this is a merged state (contact known only via a shared group, or a formerly direct contact the user removed) that can't be told apart from the field alone, so both cases are reported together rather than guessing which applies.

threema_messages -  created time, contact, direction, content type, text/filename/address, MIME type, size, location coordinates, call status/ID/duration, message state, read/delivered flags, and resolved quote-reply target, read from the decrypted message table joined against contacts. 

**The decryption work**

Threema encrypts threema4.db at rest with SQLCipher. Getting a readable database required two things, both reverse-engineered from the public, open-source wilzbach/threema-decrypt project and independently verified against this device's real files:

Unwrapping key.dat. Threema's key file (version-1 format, 45 bytes) stores the real SQLCipher key XOR-masked against a fixed 32-byte constant, followed by a SHA-1 checksum of the real key. I sourced the XOR constant from that project's cmg_clean.java, applied it to this device's real key.dat, and confirmed that the resulting key's SHA-1 checksum exactly matched the one stored in the file.
The SQLCipher PRAGMA syntax. The standard PRAGMA key = "x'HEXKEY'" blob syntax failed HMAC verification. The project's own decrypt.sh uses different, non-obvious quoting — PRAGMA key='x"HEXKEY"' — with kdf_iter = 1. (I traced this down to SQLCipher's own source: that quoting doesn't match SQLCipher's raw-hex-key pattern at all, so it falls through to treating the whole quoted string as a PBKDF2 passphrase — which is why kdf_iter matters here.)

This only covers the case where no local app passphrase was set on top of key.dat. key.dat's own protection flag is read directly rather than assumed, and the module leaves the database undecrypted rather than guess at a passphrase-derived key when that flag indicates protection.

Both encrypted artifacts open threema4.db strictly read-only, and read it alongside its write-ahead log (threema4.db-wal) rather than the main file alone. The paths glob now matches threema4.db* (the same pattern already used by WhatsApp.py for msgstore.db*), so ALEAPP extracts the -wal/-shm sidecars next to the main file, letting SQLite fold in anything committed but not yet checkpointed.

This isn't hypothetical: on the validated device, the main file alone was several weeks stale, and a real pair of static-location messages documented in the action sheet existed only in the still-uncheckpointed log. An earlier version of this module read only the bare threema4.db path and silently produced an incomplete, though not obviously wrong-looking, result (2 contacts / 30 messages instead of the real 3 contacts / 33 messages). Opening read-only means the original evidence file is never at risk -  confirmed by comparing its checksum before and after.

**Dependency**

Sqlcipher3 will need to be added to requirements.txt.  Sqlcipher3 (no -binary suffix) 0.6.2+ now ships real prebuilt wheels for Windows, macOS, and Linux, confirmed directly against PyPI's file listing, and is a drop-in replacement (same from sqlcipher3 import dbapi2 import, same statically-linked SQLCipher, no system libsqlcipher or admin rights required). The import is guarded with a module-level try/except, following the existing pattern in browserArtifactsViaMisterSkinnylegs.py: if the package isn't installed, these two artifacts are disabled with a clear log message rather than aborting the whole ALEAPP run.

**Validation**

Run against the real aleapp.py CLI, over a scoped copy of ch.threema.app's data including the -wal/-shm sidecar files (copied with cp -p to preserve original file timestamps):

threema_account: 1 row
threema_contacts: 3 rows
threema_messages: 33 rows, including the two location messages recovered only via the write-ahead log

Cross-checked against the device's published action sheet: message text, image filename/MIME/size, four call durations (100s/95s/105s/110s, matching the documented 1:40/1:35/1:45/1:50), the two location messages' addresses, and one quote-reply relationship all matched.

python3 -m py_compile: OK
pylint --disable=C,R: 10.00/10
admin/scripts/check_claim_language.py: 0 findings
admin/test/scripts/test_artifact_imports.py: pass
Original evidence file checksum confirmed unchanged after the read-only connection closes


abrignoni:

PR #1156, opened 2026-08-18
Added Private Photo Vault Support for Android

Gear-I:
**Scope, up front**

This recovers install/usage info, album structure, and per-file metadata (dimensions, timestamps, favourite/deleted/view-count state) for Private Photo Vault (com.enchantedcloud.photovault). It does not recover the actual photo/video content. That's not an oversight or a TODO: the app encrypts every imported file with an Android Keystore-backed key that is hardware-bound and cannot be extracted from a filesystem image, verified three independent ways (details below). The metadata alone still has real evidentiary value, including one finding worth flagging on its own: a second, empty "decoy" album created a millisecond after the real one.

**Summary**

Private Photo Vault had no existing ALEAPP coverage. Validated against a real Pixel 7a / Android 14 extraction (Josh Hickman's public forensic training image) and cross-checked against that image's own published action sheet, which documents a PIN being set and three pictures imported on this device.

**Artifacts**

private_photo_vault_account  - install date and the app's own usage counters (imported-media count, minutes spent in app, a key-event counter/timestamp), read from its plaintext-XML preferences file. importedMedia=3 matches the action sheet's "Imported three pictures" exactly, and installDate converts to the exact install time the action sheet documents (17:33). "Key Event Count"/"Last Key Event" are reported as raw values rather than interpreted — I don't have public documentation of what triggers them, only that the one recorded event on this device lands within a second of the media import timestamps, which is suggestive but not proof of what it represents.

private_photo_vault_albums - albums recorded in the app's ppv.db. Surfaces something worth an examiner's attention: alongside the one album holding the three real imported pictures (bucket_id = albums), a second, empty album was created one millisecond later with bucket_id = albums_decoy. Reported factually (the literal stored value) rather than asserting which app feature it belongs to.

private_photo_vault_media -  each imported file's dimensions, import time, MIME type, and favourite/deleted/view-count state, also from ppv.db.

Why the photos themselves aren't recoverable (verified three ways, not assumed)

The action sheet says: "Key used to encrypt some of the app data is not available in the extraction." I verified this independently rather than taking it on faith:

PIL/Pillow can't identify the files as any image format. Tried against both an original and its thumbnail — UnidentifiedImageError on both.
The database's stored per-file encryption_key doesn't work as a key. It decodes to 48 bytes, not a bare 32-byte AES-256 key. I tried it directly against the ciphertext with AES-256-GCM (using the file's own stored IV) two ways — as the first 32 bytes and as the last 32 bytes of the decoded value — both fail MAC verification. The 48-byte length is itself the signature of an AES-GCM-wrapped value (32-byte key + 16-byte auth tag): what's stored in the database is an encrypted key, not a usable one.
usesEncryptedPreferencesV2=true, reported by private_photo_vault_account, confirms the app uses Android Keystore-backed encryption for its sensitive data. The unwrapping key is hardware-bound (TEE/StrongBox) and never leaves the device, so no filesystem extraction can recover it.

private_photo_vault_media still reports the wrapped key and IV exactly as stored (real, verifiable data, in case a future extraction method recovers the missing unwrapping key), but this module makes no attempt to decrypt the media, and the report does not include an image element for it, since there is no valid image to embed.

**Dependency**

None. ppv.db is a plain (non-encrypted) SQLite database — read with the project's existing get_sqlite_db_records/open_sqlite_db_readonly helpers, opened read-only alongside its write-ahead log (ppv.db-wal) via a ppv.db* glob, the same pattern already used by WhatsApp.py for msgstore.db.

**Validation**

Run against the real aleapp.py CLI, over a scoped copy of com.enchantedcloud.photovault's data (including the ppv.db-wal/-shm sidecar files):

private_photo_vault_account: 1 row
private_photo_vault_albums: 2 rows
private_photo_vault_media: 3 rows, matching the action sheet's "Imported three pictures" exactly
python3 -m py_compile: OK
pylint --disable=C,R: 10.00/10
admin/scripts/check_claim_language.py: 0 findings
admin/test/scripts/test_artifact_imports.py: pass
Original evidence file checksum confirmed unchanged after the read-only connection closes


abrignoni:

PR #1185, opened 2026-08-20
cryptocurrency wallet support

CH-CLARK:
Support for four popular cryptocurrency wallets, detailed below:

- Atomic Wallet (User associated addresses)
- Bitcoin.com (User associated addresses and wallet mnemonics)
- Metamask (User associated addresses and contacts)
- Moonpay (User associated addresses)


abrignoni:
Thanks for the contribution!

- `atomic_wallet.py`: please include a fixture with this PR.
- `bitcoincom.py`: please include a fixture with this PR.
- `moonpay.py`: please include a fixture with this PR.

**Adding a fixture**

Size rules:

PR #1196, opened 2026-08-21
Add Blink App Support for Android

frahmco:
# Add Blink App Support for Android

This module adds support for **4 Blink artifacts**:

- **Camera Information**
- **Camera Entitlements**
- **Syncmodule Information**
- **Syncmodule Entitlements**

All artifacts are present in the **BlinkRoom** database.

### Test Device

This data was collected from a **Samsung SM-A166U (A16 5G)** running **Android 16** with the **2025-10-01 security patch**.

- [x] Ran the tool against a real extraction and confirmed the row counts, not just that it imports.
- [x] Ran `python admin/scripts/check_artifact_output.py <report folder>` on that report and **fixed or documented every finding**. An empty or constant column is often a real result (no group chats, coarse location denied); the fix for those is to say so in the artifact's `notes`, which is also what stops the checker reporting them.
--_Syncmodule Entitlements holds constant values on 4 columns. This is expected._
- [x] Checked it against a second app data directory (a second Android user, or two iOS containers). `--compare <multi-container report>` reads the scaling for you: it should be exactly double.
- [x] `notes`, `description` and `sample_data` say only what the data shows, and the numbers were re-derived from the finished run.

I cancelled the original PR for this app because it was poorly done. And this is my first PR and I am a git newcomer.


abrignoni:
Thanks for the contribution!

- `BlinkApp.py`: please include a fixture with this PR.

**Adding a fixture**

Size rules:

frahmco:
I may be able to use the helper. Let me verify there is no sensitive data. If that isn't possible, I will likely make a copy and sanitize the real db file. Thanks!

frahmco:
I edited data that was sensitive from the original BlinkRoom db file, nothing is particularly identifying but just to be cautious. I've attached it here, it is not very large. 

I hope to eventually get my hands on some Blink cameras to be able to control actual tests, there's much more data that could be gleaned from the app, I just don't have a way of testing without cameras to connect. Thank you for your time!

[BlinkRoom.zip](https://github.com/user-attachments/files/31424451/BlinkRoom.zip)

abrignoni:
Merged. Thank you for this, Christian. Good first contribution, and you picked an app nobody
had parsed yet, which is the hard part.

I pushed one commit to your branch before merging rather than sending you back round, so the
PR is still yours. Short version of what changed:

The `entitlement` table's primary key is `(target, target_id, name)`, so `target_id` on its
own does not name a device. Both entitlement artifacts joined on it alone, which lets a camera
pick up a sync module's entitlements if those two ids ever match. The other half of the key
was already sitting in the row: the camera's own type column carries the same value as
`entitlement.target`. Adding that returned the identical 26 and 4 rows on your file, so it
costs nothing.

The artifacts also read only the first `BlinkRoom` handed to them, and an extraction usually
carries the same app directory under two or three spellings plus a copy per Android user. They
go through `unique_files` now.

The rest is smaller. The path pattern picks up the `-wal` sidecar, the timestamps are declared
`datetime` so they sort and filter in LAVA, the two sync module artifacts no longer give one
header to two different columns, and all four carry notes. One tip from that last part:
`check_artifact_output` stops flagging a uniform column once the artifact's notes name it, so
saying it was uniform and why it still earns its place is the fix rather than dropping the
column. Handy to know for the next one.

If you do get cameras to work with, there is more in that database than we are reading.
`network` holds the account's time zone, and `key_value_pair`, `subscription`, `message` and
`tracking_event` all had rows on your sample too. Any of those would make a good second PR and
we would be happy to take it.

Thanks again.

abrignoni:
Quick follow up, Christian.

The BlinkRoom you attached is now committed as the module's test case, so the parsers stay
guarded after merge. I sanitized it first: renumbered the device, network and account ids,
swapped the hardware serials for same-length synthetic ones, and replaced the camera and
network names. I left the timestamps, the time zone and the type codes alone, since the
artifacts' notes measure those. The file was rebuilt and then checked byte by byte for
anything original. Nothing left.

The module grew while I was in there. Your four tables are now nine artifacts, covering
every table in that database that held rows: network, key_value_pair, subscription, message
and tracking_event. Each one's notes say what was measured and what was not established.

If you would rather the file was not in the repo, say the word and I will pull it.

Thanks again, for the parsers and for sending the sample.

frahmco:
Wow! Thank you very much for your time, review, and edits. I appreciate the time spent explaining the issues I had, this is definitely a learning experience for me. This started off as a small project on my own, as we had several back to back cases involving Blink cameras and information related to them. 

I hope to be able to contribute more in the future! I've told our people to keep an eye out for cameras, there is a lot more info that can be acquired if I could validate and test. I also have an iOS extraction with the app, and I am working on that one (time permitting) as well. I will be sure to improve my PR for that one when I am done with it.

As for the test file, thank you for sanitizing it completely. I am more than comfortable leaving it in the repo. 

I appreciate all of your work and look forward to contributing in the future!

PR #1200, opened 2026-08-24
Artifacts/copilot

dollyaswin:
## What this changes
Adds a new artifact parser for the Microsoft Copilot App (`com.microsoft.copilot`). It extracts user account details, conversation sessions, and individual chat messages (including AI-provided citations) from the app's offline cache and shared preferences.

- [ ] Ran the tool against a real extraction and confirmed the row counts, not just that it imports.
- [ ] Ran `python admin/scripts/check_artifact_output.py <report folder>` on that report and **fixed or documented every finding**. An empty or constant column is often a real result (no group chats, coarse location denied); the fix for those is to say so in the artifact's `notes`, which is also what stops the checker reporting them.
- [ ] Checked it against a second app data directory (a second Android user, or two iOS containers). `--compare <multi-container report>` reads the scaling for you: it should be exactly double.
- [ ] `notes`, `description` and `sample_data` say only what the data shows, and the numbers were re-derived from the finished run.

- The parser extracts messages and session details from the `offline_cache` JSON files (`offline_sessions.json` and `offline_conv_*.json`).
- Account information is sourced from the `com.microsoft.oneauth.accounts.xml` shared preferences file, which includes parsing nested JSON objects to retrieve token expiration and issue dates.
- For conversations, the parser distinguishes between `human` and `ai` authors and successfully handles extracting citation URLs and titles included in AI responses.
- Artifacts are properly namespaced (e.g., `Copilot - Sessions`) to prevent TSV/HTML file collisions with other generic app parsers.
- Empty or constant values from the Sessions cache (e.g., `Group`, `URL`, `Is Pinned`) are expected behavior and have been documented in the artifact's `notes` block to ensure they pass the artifact checker.


abrignoni:
Thanks for the contribution!

- `copilot.py`: please include a fixture with this PR.

**Adding a fixture**

Size rules:

PR #1217, opened 2026-08-26
Add the Shortcut section to the Dumpsys Parser

prosch88:
## What this changes

This adds the Shortcut Section to the Dumpsys Parser.

In addition to the usual system shortcuts (open camera, etc.), chat apps sometimes display recent or favorite contacts here. An advanced logic has been implemented for Discord. Here, parts of recent messages are also displayed below the shortcuts.

Bildschirmfoto_20260826_115611

- [x ] Ran the tool against a real extraction and confirmed the row counts, not just that it imports.


abrignoni:
Thanks for this one, the Discord message recovery is good stuff. Two issues from testing it, both confirmed with a real run:

1. A shortcut can be dumped with `intents=null` (AOSP prints that line even when there are no intents). `clean()` turns that into None and the Discord branch then calls `re.search` on it, which crashes the whole artifact. Every shortcut row is lost, not just the Discord one. A one line guard fixes it: `if package_name == "com.discord" and intents:`.

2. The dynamic sub-artifact runs a second `artifact_processor` inside the first. The decorator clears the shared Context on every call and the parent reads it back after your function returns, so on any device that has Discord shortcuts the main "Dumpsys - Shortcuts" entry in the LAVA manifest ends up with the Discord description and a blank author. The sub-artifact also can't be selected by profiles or seen by the docs generator.

Suggestion: instead of registering the Discord table at runtime, make it its own declared artifact in `__artifacts_v2__`, like the other alex_live entries, with a matching top-level function. Both functions can read the same parsed shortcut data through a module global, the way `_DUMPSYS_DICT` already works. Less code and both issues go away. I have a small synthetic dumpsys file that reproduces both and would also work as the fixture the bot asked for. Happy to share it.

prosch88:
I just saw the reply. I must have missed the notification. Thanks for the feedback. The idea behind the sub-artifact is that it allows you to add any extensions you want, but you still only need to select “Dumpsys - Shortcut.” But yes, it could also simply be a newly defined artifact. It would be very helpful if you could attach your synthetic Dumpsys file.

abrignoni:
Here is the synthetic dumpsys, with the generator so you can extend it. I put it on a branch rather than attaching it, so it keeps its history: https://github.com/abrignoni/ALEAPP/tree/fixture/alex-shortcut-synthetic-dumpsys/admin/test/fixtures/alex_live_data_shortcuts

Six shortcuts. Three Discord rows cover the `message_content` path, the `body` fallback and the `scheduled_at` fallback, one carries `intents=null` so the guard you added stays covered, and two are ordinary system shortcuts. On your current head that is 6 rows for Dumpsys - Shortcuts and 3 for the Discord table. Structure follows AOSP `dumpsys.cpp` and `ShortcutInfo.toDumpString()`. The payload inside `intents=` I could not source, so I shaped it to match your regexes. Worth checking against a real dump.

The second file is the same thing with no shortcut section. It turns up two crashes:

1. In `alex_live_shortcut` the `data_headers` and the `return` sit inside the `else`, so that branch returns None and `artifact_processor` raises `TypeError: cannot unpack non-iterable NoneType object`.

2. `alex_live_discord_shortcut` raises before it can log:

```python
if acc_dump is None and "packageName=com.discord" in acc_dump:
```

`and` still evaluates the right side when the left one is true, so `"packageName=com.discord" in None` runs. Written as `or ... not in ...` it also makes that log line reachable, which it is not today.

`alex_live_companiondevice` returns in both branches and is the shape to copy.

One note before you cut the test case. `paths` here is `('*/extra/dumpsys_*.txt')` with no trailing comma, so it is a string, and `make_test_data.py` used to walk it a character at a time. The leading `*` matched everything, so every artifact zip took the whole extraction. Pull main before you generate: #1273 landed today and fixes it. Measured on your branch with a five member zip, Shortcuts and App Ops each took all five members before, and one file each after.

abrignoni:
Merged, thank you. The Shortcuts section is a good addition to the dumpsys parser, and pulling recent Discord messages out of the shortcut intents is a nice find. Appreciate you sticking with it through the rounds.

PR #1218, opened 2026-08-26
Add Samsung Wallet enrolled-cards and receipts artifacts

Gear-I:
## What this changes
Adds two new artifacts, samsung_wallet_cards and samsung_wallet_receipts, parsing spay.db (Samsung Wallet / formerly Samsung Pay) for enrolled-card metadata and linked receipts.

 - Ran the tool against a real extraction and confirmed the row counts, not just that it imports. — Full aleapp.py run against a real Galaxy S21 (SM-G991U, Android 11) filesystem extraction; both artifacts found their 1 real row each, matching a manual harness run against the same data.
 - Ran python admin/scripts/check_artifact_output.py <report folder> on that report and fixed or documented every finding.       - First pass flagged 4 empty columns in samsung_wallet_receipts not explicitly named in the notes; fixed by naming all six empty columns (Merchant Country Code, Merchant Town, Booking Date, Value Date, Sender IBAN, Receiver IBAN) by their exact display names. Re-run reports nothing to look at.
 - Checked it against a second app data directory where the platform provides one. — Not done. The device had only one Android user profile and one spay.db; there's no second container to compare against yet, so the multi-profile loop in find_all is untested against real data.
 - Notes, description and sample_data say only what the data shows, and the numbers were re-derived from the finished run. - Sample_data ("1 row" for both) matches the actual run's output exactly.

Both tables are mostly encrypted on this device (base64 ciphertext, no recoverable key) — documented per-field in each artifact's notes rather than guessed at. The one join relationship that isn't obvious from column names: receipt.tokenID actually holds the card's enrollmentID, not its own tokenID, confirmed by direct comparison, not assumed, and reported as "Card Enrollment ID." This has only been validated against one device/one card/one receipt; no multi-card, multi-user-profile, or non-US extraction yet, so treat the encrypted-field list and the six-NULL-columns finding as single-sample until a second Samsung Wallet extraction confirms or contradicts them.

Co-Authored by: Claude Sonnet 5


abrignoni:
Thanks for this. The artifacts are solid, and being straight in the PR body about what you had and had not checked made the review a lot easier.

I pushed a commit to your branch rather than sending back a long list. Happy to talk through any of it, and say so if you disagree.

I ran the module against four registered corpora that carry `spay.db`, plus a multi-container tree: both the `/data/data` and `/data/user/0` spellings of user 0, a second Android user at `/data/user/10` with a known row delta, and a decoy package. Your scoping is correct. The duplicate storage views collapse and the second user's rows are added, so the multi-profile loop you flagged as untested does work.

Two things the extra corpora turned up.

**The receipt schema drifts.** On the Android 13 extraction the `receipt` table has no `isWatchTransaction` column, so the whole SELECT failed with `no such column` and the artifact returned nothing. That table is empty on that device so no data was lost, but on a device with receipts and that schema it would report zero rows and the framework banner would stay clean. Queries now go through `null_absent_columns`, which compiles each one against the database it will run on and reports an absent column empty instead of losing every row.

**The receipt to card link is declared in the schema.** From `receipt`'s own CREATE TABLE:

```sql
FOREIGN KEY(tokenID) REFERENCES card(enrollmentID) ON DELETE CASCADE
```

Your conclusion was right. It is worth citing the constraint rather than the value comparison, because with one card and one receipt those two columns compare equal whatever the real relationship is.

Smaller ones:

- `source_paths` was appended inside the row loop, so the report's "located at" line repeated a database once per row it contributed. Invisible at one row, four copies of one path on a four row run. Now appended once per file.
- `check_artifact_output.py` still reported four empty columns. The notes named them as "merchant country/town" and "sender/receiver IBAN", and the checker matches the exact display name, so only two of the six cleared and Value Date was not named at all. All six now use their exact names.
- `cardHolderName` and `cardBalance` are empty strings rather than ciphertext on that extraction. The rest of the encrypted list checks out.
- Enrolled Cards now leads with Card State Updated.
- `sample_data` keys the corpus as `cookbook_a11`, which is what your own test case file names, so `validate_sample_data --registry` can resolve it. Added the three corpora that hold `spay.db` with no card or receipt rows, and left the Android version off one of them because its extraction has no readable build.prop.

New artifact, `samsung_wallet_card_art`. `cardArt.cardArtUri` records each image's on-device path and `cardArt.enrollmentId` names the card it belongs to, so images join to cards from the database rather than by file name or by matching size and timestamp. The index is keyed on the app container as well as the file name, so one Android user's images cannot be joined to another user's rows. On the tested extraction that is four PNGs per card including the issuer logo, which earns its place precisely because the card's own name, brand and last four digits are encrypted.

Test data re-cut for the new artifact and the column order change.

PR #1228, opened 2026-08-28
Recover superseded records from SQLite write ahead logs and journals

abrignoni:
walStrings runs a printable-character regex over every -wal and -journal file and links the result. That reads a structured store as text: it decodes the bytes as UTF-8 and drops what will not decode, it cannot say which table or column a string came from, and it reports one row per file rather than per record.

Both file types describe themselves. A write ahead log is a 32 byte header followed by frames, each a 24 byte header giving the page number and the database size at its commit, then one complete page image. Writing those images onto a copy of the database at the offsets their own headers name produces a valid database, and SQLite decodes the values from the schema the database carries. Nothing searches for record shaped bytes.

Two artifacts share one pass:

- SQLite Records Superseded in a WAL or Journal: every row that appears in a rebuilt state and is absent from the current one.
- SQLite Journals Holding Superseded Records: one row per database with the frame counts, so the databases worth opening can be found without reading the first table end to end.

Measured on four extractions: 8354, 9712, 11931 and 29375 records, the slowest run taking 76 seconds. A controlled database with 40 rows inserted and 13 deleted returned exactly those 13, with no misses and nothing spurious.

A recovered row means it was present at an earlier commit and is not present now. It does not establish who removed it: an app trimming a rolling table gives the same result as a deletion. Most of what came back on the tested images is application and framework bookkeeping, and no message table was recovered on any of them. Frames from an earlier log generation are counted and not rebuilt. Every per database limit logs the database and what was skipped. A sidecar with no database beside it is counted rather than guessed at.

walStrings is unchanged: it still covers the files that hold no parsable structure.

### DIFF
--- a/scripts/artifacts/walRecords.py
+++ b/scripts/artifacts/walRecords.py
@@ -1,0 +1,378 @@
+__artifacts_v2__ = {
+    "get_walRecords": {
+        "name": "SQLite Records Superseded in a WAL or Journal",
+        "description": "Rebuilds each committed state a SQLite write ahead log or rollback "
+                       "journal records, and reports the rows present in one of those states "
+                       "that the current state of the database no longer holds.",
+        "author": "@AlexisBrignoni, Claude",
+        "creation_date": "2026-08-28",
+        "last_update_date": "2026-08-28",
+        "requirements": "none",
+        "category": "SQLite Journaling",
+        "notes": "A write ahead log stores complete page images, each carrying the page number "
+                 "and the database size at the commit it belongs to, so the state after any "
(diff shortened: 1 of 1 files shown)


abrignoni:
Closing this. The output is hard to navigate: one generic table spanning every database on the device, most of it framework bookkeeping, which leaves the examiner to work out what matters.

Recovery is better done per database inside the artifact that already parses it, so the recovered rows come out in the same columns as the live ones and can be read the same way.

The branch stays up. The frame parsing and the state rebuild are the reusable parts.

PR #1229, opened 2026-08-28
Create untappd.py

stark4n6:
## What this changes

1. Adds 7 parsers for the Untappd app.

- App Events
- Cached Checkins
- Checkin Location Suggestions
- Device Analytics
- Discover Locations
- Push Notifications
- User Profile

2. Update and aligns author name across all affected parsers

- [x] Ran the tool against a real extraction and confirmed the row counts, not just that it imports.
- [x] Ran `python admin/scripts/check_artifact_output.py <report folder>` on that report and **fixed or documented every finding**. An empty or constant column is often a real result (no group chats, coarse location denied); the fix for those is to say so in the artifact's `notes`, which is also what stops the checker reporting them.
- [x] Checked it against a second app data directory where the platform provides one (a second user, account, or container). `--compare <multi-container report>` reads the scaling for you: it should be exactly double.
- [ ] `notes`, `description` and `sample_data` say only what the data shows, and the numbers were re-derived from the finished run.


abrignoni (commented):
Thanks Kevin. The author cleanup across the 49 modules is good and reads clean, that half is ready as is. The Untappd module needs work first.

I ran it against a constructed Untappd tree. Three of the seven artifacts do not run at all.

## Blocking

**1. `source_paths` is never defined.** Lines 472, 500 and 597 return `'\n'.join(sorted(source_paths))`. Cached Checkins, Discover Locations and Checkin Location Suggestions each die with `NameError: name 'source_paths' is not defined` and report zero rows on any device that has the app. That is the E0602 the lint job is red on. The run still prints "Report generation Completed" and exits 0, so it is easy to miss.

**2. `untappd_notifications` returns a bare `None`** when the XML holds no notification entries, and the wrapper unpacks the return value, so that path raises `TypeError: cannot unpack non-iterable NoneType object`. That is the normal state of a fresh install. Return `data_headers, [], source_path` instead.

**3. Multi-container scaling is off in both directions.** Same container under `/data/data` and `/data/user/0`, a second user at `/data/user/10`, plus a decoy package:

| artifact | one container | multi | expected |
|---|---:|---:|---:|
| Cached Checkins | 1 | 4 | 3 |
| Discover Locations | 1 | 3 | 2 |
| Checkin Location Suggestions | 2 | 6 | 4 |
| App Events | 5 | 5 | 7 |
| User Profile | 1 | 1 | 2 |
| Device Analytics | 1 | 1 | 2 |
| Push Notifications | 1 | 1 | 2 |

The three loop artifacts iterate `context.get_files_found()` raw, so one checkin comes out once per storage spelling. `unique_files` from `storagePathViews` fixes those three exactly, I measured 4, 3 and 6 going to 3, 2 and 4 with both users kept. The other four use `get_file_path`, which returns the first match and stops, so a second Android user's data gets staged and then dropped. Those want a loop over the containers.

The decoy contributed zero rows to all seven, so your paths are scoped correctly.

## Before merge

**4. Five of the seven descriptions have words run together** from implicit concatenation with no trailing space: "Untappd appsuch as model", "app statussuch as launch", "beer infoas well as", "locations forfeeding local events", "locationssuch as recent". Those reach the report and the LAVA manifest.

**5. `notes` is empty and `sample_data` is missing on all seven.** Two worth writing down: the Foursquare rows are server returned nearby venues rather than places anyone went, and the notification timestamp is read as microseconds out of the FCM key.

**6. `output_types` of `"standard"` skips KML** (`ilapfuncs.py:155`). Cached Checkins, Discover Locations and Checkin Location Suggestions all carry lat and lng, so those three want `"all"`.

**7. `strptime` parses `%z` and `strftime` then drops it**, in Cached Checkins and Checkin Location Suggestions. A `-0400` checkin renders as 17:08:15 when the UTC value is 21:08:15. If Untappd always returns +0000, say so in the notes. If it does not, the datetime column is off by the offset.

## Smaller

8. `Distance (Miles)` states a unit. Worth a line in the notes on where that comes from.
9. `where name IS "device_attributes"` uses double quotes. SQLite reads those as an identifier first and only falls back to a string literal, so single quotes are safer.
10. Rest of the lint: unused `convert_unix_ts_to_utc` import, unused `beer_abv` (looks like it was meant to be a column), unused `source_name` in `untappd_profile`, three f-strings with nothing interpolated, three bare `except Exception`. `untappd_discover_locations` is also the only one of the seven not calling `data_list.sort()`.

Icons all resolve, the Source File columns are extraction relative with no local paths in them, and the `.0` sibling lookup for the cache date works. With just the three crashes patched, the parsing produced correct rows for all seven on my tree, so this is close.

PR #1235, opened 2026-08-28
Add Google Play Movies & TV signed-in account artifact

Gear-I:
## What this changes

Adds `scripts/artifacts/googlePlayMovies.py` with one artifact, `google_play_movies_account`. It reads `com.google.android.videos`'s `purchase_store.db` and reports each Google account known to the app (`user_data.user_account`) alongside whether that account currently has a synced configuration from Google's Play backend (`user_configuration.config_account`), surfaced as a "Play Config Synced" Yes/No column.

- [x] Tested against a real extraction, not synthetic sample data. Validated against two real devices: a Samsung A53 SM-S536DL (Android 14, Hexordia/Magnet Virtual Summit CTF 2026 image) and a Google Pixel 7a (Android 14, Josh Hickman public image). Ran the full `aleapp.py` pipeline against both, and against a combined fixture with both accounts in one report.
## Validated results:
 "pixel7a_a14": "Android 14 | com.google.android.videos | 1 row",
 "samsunga53_a14": "Android 14 | com.google.android.videos | 2 rows",

- [x] Ran `admin/scripts/check_artifact_output.py` against the report output and fixed anything it flagged. Clean on all three runs — no empty/constant/identical-column or sparse-lead findings.
- [x] Checked a second app data directory / multi-account case on the test device. The A53 image genuinely has two Google accounts registered with this app (`danieljonesonline55@gmail.com` and `djones.iss.secure@gmail.com`), and only one has a config row — confirmed by direct comparison of both tables, and the artifact correctly reports both states in the same run.
- [x] Notes/description reflect what was actually observed, not assumed. The notes are explicit that `purchased_assets`, `search_history`, `wishlist` and `watch_next_feed` were empty on both real devices despite the database carrying substantial WAL activity — that activity turned out to be the app re-syncing static reference data, not purchase or watch history, so the artifact reports identity only and says so.

This is a deliberately thin artifact,  I went in expecting purchase/rental history given the WAL sizes and confirmed there isn't any recoverable here, so I scoped it down to what's real rather than pad it with unpopulated columns. `sync_snapshot_token`, `config_play_country` and `account_links` were NULL/0 on every real row checked and aren't reported; `wishlist_snapshot_token` held the identical value across every account on both devices, which reads as a static default rather than a real sync cursor, so it's excluded too, all documented in the notes rather than silently dropped.

Co-Author: Claude


abrignoni:
## Review of the submitted artifact, and what this change fixes

The parser is sound and I would rather correct it here than send it back, so this
lands the fixes directly. This note records what changed and why, so the reasoning
sits with the code.

### What the module already got right

- The trailing `*` on the path pattern is load bearing and it is there. On the Samsung
  A53 image the `.db` file is 4,096 bytes and reads as zero tables without its 267 KB
  write ahead log. Every row in that fixture lives in the log.
- Sidecars and matched directories are excluded before anything is opened, so a
  `-wal` can never end up cited as the source of a row.
- The source path is set from the files that actually returned rows, not from the
  files that were attempted.
- The case1 fixture carries all three storage views of the same database, so CI
  genuinely exercises the duplicate view collapse instead of asserting it.

### Claims the store does not support

Each of these was written as an observation. None of them is one.

1. **"lists every Google account this app instance has ever been used with."**
   `user_data` has three columns and no timestamp. Nothing in it records use, and
   nothing establishes that the list is complete. What is provable is that the table
   holds one row per account, keyed on the account name.

2. **"means that account currently has an active, synced configuration from Google's
   Play backend."** What is provable is that a row exists in `user_configuration` for
   that account and that the row carries a configuration blob. "Currently", "active"
   and "synced" are not in the data and no published source was cited for them.

3. **"evidence that a second account was added to the app without becoming (or after
   ceasing to be) the one actively used."** This is a statement about what a person did,
   derived from a row that is not present. An absent configuration row has several
   ordinary explanations and the store distinguishes none of them. This is the one that
   had to go: it is a finding about a person, in a field that reaches the report and the
   LAVA manifest, and it would have been quoted in casework.

4. **"it reads as a static default baked into the app."** The observation is that
   `wishlist_snapshot_token` held the same value on every account across both images.
   That is worth stating. Where the value comes from is an inference and is now written
   as one.

5. **"A second Android user profile has its own copy of `purchase_store.db`, so every
   copy found is read."** The code does this and the code is correct. Neither validated
   image has a second Android user, so the behaviour was never exercised. Two images in
   our corpus do have one, each carrying a different account under `/data/user/10`, and
   the artifact now records that it was tested against them.

6. **The distinction the artifact exists to surface appears once.** Across the sixteen
   Android images that carry this database, nineteen accounts are listed and eighteen of
   them have a configuration row. One account, on the image this artifact was built
   from, does not. One observation is enough to report the state. It is not enough to
   say what the state means, and the notes now say so.

### What was not examined

- **`user_configuration.config_proto`.** This is the only populated blob in the table
  the artifact already queries, 1,702 bytes on both validated images, and it was not
  named anywhere. Decoding it with the vendored blackboxprotobuf confirms it is Google's
  static reference configuration: TVPG and MPAA rating systems, `//play.google.com/`,
  `//android.clients.google.com/`, `familysharingenabled`, `countries_with_tv`.
- **The account storefront country is available after all.** The notes said
  `config_play_country` was NULL on every account, which is true, and concluded the
  country was not recoverable, which is not. It is the first field of `config_proto`.
- **`guide_settings`.** A third table keyed on the account, one row on the Pixel 7a,
  never named. Naming a table and saying it was excluded is what stops the next person
  re-deriving the same dead end. An omission and a reasoned exclusion look identical in
  a diff.
- **A guard on an unreadable `user_configuration`.** The helper returns an empty result
  on a SQLite error, so a missing or unreadable table would have printed "No" for every
  account, with the error visible only in the log. The table is present on all sixteen
  images, so no image reaches that branch. The guard added here was verified against a
  constructed copy with the table removed: the account is still reported, the value is
  left blank rather than No, and the run log carries the reason and the path. A wrong
  value in a report is worse than a missing row.
- **Attribution across profiles.** Where two Android users each hold an account, the
  submitted output gives two rows with nothing saying which profile each belongs to. For
  an identity artifact, work profile versus primary is the finding.

### Why this needed a second pass

This artifact was written with model assistance, and that is fine. What is not fine is
shipping what the model wrote without running the queries behind it.

The submitted notes read as thorough. They name tables, they name columns, they say what
was checked and what was found empty. That texture is not evidence of anything. The proof
is in a single paragraph of the submitted text, where two claims sit side by side. One
says the write ahead log traffic is the app re-syncing static reference data, content
rating systems and API endpoints. That claim is correct, and decoding `config_proto`
confirms it line for line. The other says a missing configuration row is evidence about
how an account was used. Nothing supports it. Same paragraph, same confident voice, and
the only thing that separates them is running the query.

Green CI does not help here. All ten checks passed on the submission. The claim language
gate reads `name` and `description` only, so `notes` goes unread, and `notes` is where the
heavy claims were.

Checking was not expensive. Sweeping every registered Android image took a few minutes and
changed the picture: sixteen carry this database rather than the two that were tested, two
of them have the second Android user the notes claimed to handle, and the distinction the
artifact reports turns out to appear on one account in nineteen.

The standard for anything in `description`, `notes` or `sample_data` is that a person can
point at the query that produced it. If a sentence cannot be traced to a query someone ran,
it either comes out or it gets labelled as an inference. That applies to text a model wrote
and to text I wrote, equally.

For future submissions, two things make this quick to check: say which claims are proven
against real data and which are inferred, and name every table in the store with the side
of the line it falls on, the excluded ones included.

PR #1289, opened 2026-09-01
Cryptocurrency Wallet Support & Test Data

CH-CLARK:
## What this changes
Support for four popular cryptocurrency wallets on Android and test data:
- Atomic Wallet
- Bitcoin.com
- MetaMask
- Moonpay


abrignoni:
Thanks for this, CH-CLARK. Four wallets in one PR is a real contribution and the parsing logic is sound. I ran all six artifacts against your own fixtures and got 97, 8, 3, 5, 5 and 4 rows. Good stuff. A few things to sort before it can go in.

**Two artifacts crash.** bitcoincom.py and moonpay.py assign inside the file loop and read outside it, so when the glob matches only the sidecars nothing is bound:

    UnboundLocalError: cannot access local variable 'all_rows'

Reproduced on both. The moonpay one is worse, because your `except MMKVError: continue` is exactly what routes an encrypted or malformed store into the crash.

**metamask.py opens whatever came last.** In both functions the for body is only the `if not ... continue`, and `source_path = file_found` plus the `open()` sit outside the loop. The filter does nothing. Pylint flags it twice as W0631.

**All six drop a second Android user.** Every module reads one file and stops. I re-rooted your fixtures under `data/data` and ran the repo's own gate:

    python3 admin/test/scripts/run_test_cases.py --storage-views \
      --module atomic_wallet --module bitcoincom --module metamask --module moonpay

All six second-tenant legs fail at 1x where 2x is required. On a device with a work profile, one user's addresses, contacts and seed phrases go missing from the report with no error. Loop over every match and accumulate.

That gate never got a chance to tell you, because your fixtures sit at `<App>_TestData/<package>/...` with no `data/data` or `data/user/0` segment, so it skips all twelve legs. Please re-cut them at a real extraction path.

**CI.** Test Cases fails: six units report NO_BASELINE, exit 1. Record them with `test_module.py`. The four case JSONs also carry `expected_output` empty while those same fixtures produce rows. Pylint fails too: missing final newline on all four files, trailing whitespace, long lines, `open()` with no encoding.

**Smaller items:**

- atomic_wallet never assigns `source_path`, so a 97 row run returns `''` and the located-at line is empty. The other three set it to the last file read only.
- atomic_wallet computes `batch.timestamp` and discards it. Approximate, but it is the only time signal there and the artifact has no time column.
- metamask uses `details['memo']` where the sibling field uses `.get`. One entry without a memo kills the artifact.
- moonpay uses `print(e)`. Use `logfunc` so it reaches the run log.
- Four of six artifacts have empty notes, including the mnemonics one. That artifact publishes wallet seed phrases and says nothing about what they are or what you verified.
- The bitcoincom note sends examiners to a live third-party site to derive addresses from an XPUB. Pasting an XPUB into a website hands that site the whole wallet's address history. Please drop that line.
- `requirements` says none, but these need ccl_chromium_reader and scripts/mmkv_parser. The first only resolves transitively through mister_skinnylegs.
- `Cryptocurrency` is a new category here. Fine by me, just flagging it.
- `('mnemonic',)` should be title case like the rest.

`currency-bitcoin` checks out, it is in the bundled Tabler CSS.

Send the updates over when you have them and we will be happy to take another look.

abrignoni:
One more thing I should have flagged the first time, and it is the one that cannot be undone after a merge.

The Bitcoin.com fixture commits two 12-word BIP39-shaped seed phrases in plaintext, inside `admin/test/cases/data/bitcoincom/`, and every address in the other three fixtures derives from those keys. I am assuming these are throwaway wallets you made for this, which is the sensible way to do it, but please confirm. A public history cannot be recalled.

Two corrections to sharpen what I said before:

**The ccl_chromium_reader point is worse than "requirements says none".** Because the import sits at module level and the plugin loader has no try/except around module load, an absent `ccl_chromium_reader` does not skip the artifact, it stops ALEAPP from starting at all. Measured: dependency present, 1264 artifacts load; dependency absent, `PluginLoader()` raises; dependency absent with `atomic_wallet.py` removed, 1235 artifacts load. `tutanota.py` handles the same dependency with the import inside the function in a try/except, and that pattern survives it.

**The Atomic Wallet description does not hold.** "user associated cryptocurrency addresses" against 97 rows that carry only 61 distinct addresses, 2 of them blank, all from a single `addresses` record. That is the app deriving an address for every asset it supports rather than every asset the user holds, which is also why the EVM chains collapse onto one address. The same store has a separate `activeWallets` key that is not parsed, and that one would support a claim about the user's own wallets. Either parse it, or reword to what the rows are and say the row unit in the notes.

Small one while you are in there: the mnemonics artifact returns 5 rows with 2 distinct values. Worth deduplicating or saying in the notes why the repeats are there.

Issue #515, opened 2024-08-28
GUI interface doesn't work


Zspec777:
The bottom of the GUI interface window, where Process and Close buttons are located, is not accessible. I can scroll down the selection list of the modules but cannot get to the Process button. 
Any suggestions how to repair it?
Thank you

![Screenshot 2024-08-28 144508](https://github.com/user-attachments/assets/8f37bda7-ec79-4d7a-acf2-92f459c39a22)

Johann-PLW:
Hello @Zspec777,

Thank you for reporting this bug.
Could you please provide the resolution of your screen?
I think you have this issue because we had to use fixed width and height to center the window in the screen and the height is too high for your resolution.
To resolve this issue, you need to use a higher resolution.
I will have a look if I can update the code to reduce the height of the ALEAPP's window if the resolution is too low.

Zspec777:
1st - 3440 x 1440, 2nd - 1080 x 1920. Both are the highest. I already tried using only one display.

Johann-PLW:
Hello @BorntoExplore,
According to your screenshot, you are using ALEAPP on Linux.
I was able to reproduce this issue.
Have a look in your settings -> Accessibility and turn Off Large texte (sorry my Ubuntu is in French)
With Large text turned on:
![image](https://github.com/user-attachments/assets/af031845-cb40-4c0f-8d3d-07a7ab8ee430)

With Large text turned off (You need to restart ALEAPP):
![image](https://github.com/user-attachments/assets/043705a9-c420-47bf-8107-451cf22c79cd)

Let me know if it works for you.

Johann-PLW:
Hello @Zspec777 ,
It is probably related to the text size used by your Windows operating system.

Try to decrease the text size:
- On Windows 11, click on the Windows logo, Settings, Accessibility and Text size
- On Windows 10, click on the Windows logo --> Settings --> Ease of Access --> Display --> Make text bigger

Slide the cursor to the left and restart ALEAPP.
Let me know if it works for you.

jtchevy:
Hi all, I updated to latest version and ran on FFS of an Android. Windows GUI, it will run and complete. I select Close and Open Report, it only goes back to ALEAPP. I can see all the data in Exported Folder. I ran the same extraction in an older version and it worked.
Thank you.
JLT

abrignoni:
[...]

PR #1407, opened 2026-09-16
feat: add batch and PowerShell launcher scripts for CLI and GUI

sierrablue8866-droid:


abrignoni:
Hi @sierrablue8866-droid, thank you for this PR and for the time you put into the GUI.

Docking the Process bar at the bottom fixes a real problem. When the window's contents don't fit, for example with a larger system text size, the Process and Close buttons get cut off, which is what #515 reports. The processing screen had the same problem with the progress bar and the buttons shown at the end of a run, and all five LEAPPs share this layout, so we applied your approach to all of them:

- abrignoni/ALEAPP#1410
- abrignoni/iLEAPP#2208
- abrignoni/RLEAPP#479
- abrignoni/VLEAPP#220
- abrignoni/DLEAPP#180

I'm closing this one in favor of those. A few notes on the other changes, for next time:

- The launcher scripts expect a `.venv` folder at the repo root. That's a local setup rather than something the README documents, so we'll leave them out.
- We couldn't find a bug fixed by the lint and typing changes, and one of them introduces one: `button_frame` was renamed to `clear_btn_frame` in the Clear History dialog, but three of its buttons still use the old name, so they show up in the main window's module toolbar instead. Our lint check only fails on warnings a PR adds, so existing warnings don't need a file-wide disable.
- Keeping each PR to one change makes it much easier for us to review and merge.

Thanks again. Issues and PRs are always welcome.

